PR #7, opened 2018-10-20
Add a Homebrew tap release process

toolmantim:
Adds [Tap Release](https://github.com/apps/tap-release) for automatically publishing a Homebrew formula to [buildkite/homebrew-cli](https://github.com/buildkite/homebrew-cli). After this, people will be able to `brew install buildkite/cli/bk` and `brew update bk` to update.

- [x] Make this repo public
- [ ] Give this a test

p.s. the brew won't work until this repo is made public though, and neither will Tap Release.

p.p.s. it'd be amazing to have some kind of update detection early on, so it could print the command for them to update and push people along w/ new versions

p.p.p.s. I've never used p.s. in a PR before


pda:
@l-suzuki just noticed that there is a (different) core homebrew formula called `bk` now:
https://github.com/Homebrew/homebrew-core/pull/95043

So `brew install buildkite/cli/bk` still works, but things like this which used to work no longer do:

```bash
brew tap buildkite/cli
brew install bk
```

I guess people just need to be slightly more careful to use the fully qualified `buildkite/cli/bk` name now.
Or, we could rename our formula to e.g. `bk-cli`.

PR #13, opened 2018-11-17
Rename commands

lox:
I tried a slightly unusual command / sub-command ordering initially, with `<action> <subject>`, but I don't think it works well. Before we hit 1.0 I'm going to revert this to be inline with our other tools:

* `bk create build` to `bk build create`
* `bk list pipelines` to `bk pipeline list`
* `bk local run` to `bk run`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,16 +19,16 @@ bk configure
 bk init
 
 # Triggers a build via the cli
-bk create build
+bk build create
 
 # Opens the current pipeline in your browse
 bk browse
 
 # Lists pipelines that you have access to
-bk list pipelines
+bk pipelines list
 
 # Runs a build entirely locally for development
-bk run local
+bk local run
 ```
 
 ## Development
--- a/cmd/bk/main.go
+++ b/cmd/bk/main.go
@@ -223,14 +223,17 @@ func run(args []string, exit func(int)) {
 	// --------------------------
 	// run command
 
-	runCmdCtx := cli.RunCommandContext{}
-	runCmd := app.Command("run", "Run a pipeline locally").
+	localCmd := app.Command("local", "Operate your local repositories")
+
+	runCmdCtx := cli.LocalRunCommandContext{}
+	runCmd := localCmd.
+		Command("run", "Run a pipeline locally").
 		Default().
 		Action(func(c *kingpin.ParseContext) error {
 			runCmdCtx.Debug = debug
 			runCmdCtx.Keyring = keyringImpl
 			runCmdCtx.TerminalContext = &cli.Terminal{}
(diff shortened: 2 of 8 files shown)


petemounce:
I like having the `bk local <local things>` separation to make it clear that that won't affect the remote buildkite.com things.

Is it possible to dump out a tree-view in ASCII of the current command structure? We've got our own workflow tool, and we found that was helpful when figuring out which commands to put where.

Here's our sketch in a markdown:

```markdown
```
circle
  check (intend to fold this into `merge --dry-run`)
        (intend this to become presubmit)
  merge (keep this)
  
  push         (idea: wrap git push and make sure pre-push hook is installed and runs)

  ci           (idea: deal with CI)
    pipeline
      new      (idea: create new safe-defaults pipeline configuration files)
        --name <name>     // pipeline name
        --generate [steps,definition] // to generate just steps or just definition files
        --directory <dir> // into here; otherwise, current repository's .buildkite folder
      push     (idea: push a pipeline to BuildKite based on the specified config file)
        --name <name>     // pipeline name
        --directory <dir> // from here; otherwise, current repository's .buildkite folder
      pull     (idea: take a remote pipeline from BuildKite and create a local representation)
        --name <name>     // pipeline name
        --directory <dir> //to here; otherwise, current repository's .buildkite folder
      check    (idea: verify deployed pipelines correspond to source controlled configurations. Default behaviour: check all corresponding to working dir repo)
        --all             // verify all 
        --name <name>     // pipeline name
        --directory <dir> // to compare pipeline against specific configuration instead of the one in .buildkite
      delete   (idea: delete a deployed pipeline)
        --name <name>     // pipeline name

    trigger    (idea: trigger actual CI)   
    check      (idea: run what CI would, locally)
    status     (idea: has CI run and passed against HEAD?)

  local        (idea: commands delegate into by-convention bash scripts within current repo)
    prebuild   (idea: lint, fast tests pre-compilation step)
    check      (idea: alias for prebuild)
    build
    test
    publish

  review       (idea: deal with reviews)
    --id <id>  (infer from branch if possible, but allow to specify)

    ask        (idea: ask for review of branch. Specify reviewers; that might be automated before this via prow.)
    chase      (idea: ping people that are stale on given review, or ask for next round after comments processed.)
  
  init         (idea: clone a repo, and set up configured hooks once it arrives)
    --git <git url to clone from>
  
  pr
    new        (idea: alias for review-ask)

  branch
    new        (idea: way to create a branch with a semantic prefix that we can then use within release note generation, semver, etc) 
      feature
      fix
  
  version
```

lox:
I agree @petemounce! Changed to `local run`.

PR #23, opened 2019-02-05
Update README.md

aaronsky:
Homebrew didn't support the original options without tapping this repository first. I've updated the installation instructions accordingly. If there's a preferred installation strategy that nullifies this, let me know and I'll update my PR accordingly.


lox:
Thanks, you are absolutely correct!

PR #25, opened 2019-02-06
Implement text and select block steps

lox:
This provides an implementation of https://buildkite.com/docs/pipelines/block-step, using the UI components from https://github.com/manifoldco/promptui.

render1549425024351

### DIFF
--- a/local/pipeline.go
+++ b/local/pipeline.go
@@ -109,8 +109,24 @@ func (s step) String() string {
 	return "Unknown"
 }
 
+type blockField struct {
+	Text     string              `json:"text"`
+	Select   string              `json:"select"`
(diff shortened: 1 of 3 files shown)


lox:
@toolmantim the code is a mess, but I'd love feedback on the interaction design and things like spacing, etc.

toolmantim:
This is awesome! There’s not a lot to say, it looks pretty good.

Perhaps the only two things I can see are the ✔️ character seems odd for a block, and that you could probably do without the final [y/n] — perhaps just unblock as soon as they fill out the last field? (Unless it’s a field-less block step, in which case you’ll probably still need it)

lox:
I'm going to merge this for now!

PR #30, opened 2019-02-08
Add plugins-path and switch to temp dirs

toolmantim:
In #20 we added a temp file path for builds dir, but I think this was meant to be a temp directory and not a file? This fixes that, adds a plugins-path too, so that you can run pipelines that use plugins within a cleanroom environment (like a minimal Docker container).

Tested locally, and seems to work 👌🏼

### DIFF
--- a/local/run.go
+++ b/local/run.go
@@ -289,10 +289,19 @@ func (a *Agent) Run(ctx context.Context) error {
 		cmd.Stderr = ioutil.Discard
 	}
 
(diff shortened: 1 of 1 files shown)


lox:
Looks good! Good call on the defer's, I can't remember why I didn't clean them up afterwards 🤔 I clearly wrote #20 whilst I was half asleep.

toolmantim:
😁 I'm just stoked I wrote and ran some local Golang code

lox:
I am too! ❤️

PR #38, opened 2019-02-18
Fix default keychain selection

lachlancooper:
This fixes a typo; `keyring` is not a valid keyringBackend.


lox:
Ugh, I knew those not being consts would bite us eventually. Thanks and apologies.

PR #50, opened 2019-04-11
fix spelling in error messages

jsleeio:


lox:
Awkward! Need to get a spelling linter in there! Thanks for the catch @jsleeio

Issue #28, opened 2019-02-06
Local Pipeline run failing on Windows


jamiebrynes7:
Hi there, I'm attempting to run a pipeline locally on Windows, and am having some trouble. I've pasted the output below and you can find a copy of the static steps [here](https://github.com/spatialos/gdk-for-unity/blob/develop/.buildkite/premerge.steps.yaml).

Any ideas where I'm going wrong? 

Thanks

```
λ bk local run .buildkite/premerge.steps.yaml --debug
2019/02/06 18:29:18 [keyring] Considering backends: [file]
2019/02/06 18:29:18 Serving API on http://127.0.0.1:49256
2019/02/06 18:29:18 Creating bootrap script at C:\Users\JAMIEB~1\AppData\Local\Temp\bootstrap-450177923
>>> Starting local agent 🤖
>>> Starting build 👟
>>> Executing initial command: buildkite-agent pipeline upload ".buildkite/premerge.steps.yaml"

  _           _ _     _ _    _ _                                _
 | |         (_) |   | | |  (_) |                              | |
 | |__  _   _ _| | __| | | ___| |_ ___    __ _  __ _  ___ _ __ | |_
 | '_ \| | | | | |/ _` | |/ / | __/ _ \  / _` |/ _` |/ _ \ '_ \| __|
 | |_) | |_| | | | (_| |   <| | ||  __/ | (_| | (_| |  __/ | | | |_
 |_.__/ \__,_|_|_|\__,_|_|\_\_|\__\___|  \__,_|\__, |\___|_| |_|\__|
                                                __/ |
 http://buildkite.com/agent                    |___/

2019-02-06 18:29:18 NOTICE Starting buildkite-agent v3.8.4 with PID: 40328
2019-02-06 18:29:18 NOTICE The agent source code can be found here: https://github.com/buildkite/agent
2019-02-06 18:29:18 NOTICE For questions and support, email us at: hello@buildkite.com
2019-02-06 18:29:18 INFO   Configuration loaded from: C:\buildkite-agent\buildkite-agent.cfg
2019-02-06 18:29:18 DEBUG  Bootstrap command: C:\Users\JAMIEB~1\AppData\Local\Temp\bootstrap-450177923
2019-02-06 18:29:18 DEBUG  Build path: C:\Users\JAMIEB~1\AppData\Local\Temp\buildkite-build-735988486
2019-02-06 18:29:18 DEBUG  Hooks directory: C:\buildkite-agent\hooks
[...]

PR #58, opened 2019-05-19
Set environment needed for windows processes

lox:
This addresses the initial error from #28. I'm still not able to get it working locally, but making progress.


tduffield:
Couple things I noticed on this branch. 

1. We need to import `runtime` 
2. I get a different checksum for `github.com/keybase/go-keychain` that I need to manually fix in my build
3. We probably need to do another `if` block when we create the bootstrap-script to write it out as a `.bat` file. 
4. The bootstrap-script file frequently gets a "file not found" at some point in the run. When I manually hack in a static .bat bootstrap script, things work fine.

PR #59, opened 2019-05-27
Add a rotate-webhooks command

lox:
This adds a command for rotating buildkite webhook urls for all your pipelines, and then updates those webhooks in the associated github repository. 

It's working, you can try it out with:

```bash
git clone https://github.com/buildkite/cli.git
cd cli
git checkout add-webhook-rotation-command
go run ./cmd/bk configure
go run ./cmd/bk rotate-webhooks --org my-buildkite-org --pipeline my-buildkite-pipeline --dry-run
```

If you leave off the `org` or the `pipeline` it will process all the pipelines (you can use `--prompt` for a `y/n`), but I'd suggest starting off with just a few pipelines.

We'll be testing this before broader release. Feedback welcomed!

### DIFF
--- a/cmd/bk/main.go
+++ b/cmd/bk/main.go
@@ -192,6 +192,36 @@ func run(args []string, exit func(int)) {
 		StringVar(&initCtx.PipelineSlug)
 
 	// --------------------------
+	// rotate-webhooks command
+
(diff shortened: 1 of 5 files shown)


lox:
Moving this to a dedicated tool: 

https://github.com/buildkite/github-webhook-rotate

PR #61, opened 2019-06-06
Add --env flag to `bk build create`

slam:
For example,

$ bk build create --env ENV1=VAR1 --env ENV2=VAR2


lox:
Sorry I missed this!

PR #64, opened 2019-08-25
Artifact download

### LINKED ISSUE #15: Artifact download
I often want to download all artifacts for a particular build or job to my filesystem for local build pipeline debugging. It'd be neat if `bk download artifacts https://buildkite.com/acme/widgets/builds/123` worked, or something.

sj26:
Adds a `bk artifact download` command which can download artifacts from a job, or a whole build of jobs, and accepts a glob for filtering downloaded artifacts. It has some caveats — I haven't added pagination because it's impossible to paginate the artifacts connection — and globbing is done client-side for the same reason so may not be the same syntax as `buildkite-agent artifact download`. It also needs tests. But it works well enough. Fixes #15.

### DIFF
--- a/cmd/bk/main.go
+++ b/cmd/bk/main.go
@@ -287,6 +287,27 @@ func run(args []string, exit func(int)) {
 		IntVar(&pipelineListCtx.Limit)
 
 	// --------------------------
+	// artifact commands
+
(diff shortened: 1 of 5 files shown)


lox:
I rebased this and merged it into master via the CLI.

Issue #83, opened 2020-05-16
brew failed: Download failed: $STABLE_URL


drnic:
```
$ brew upgrade
...
==> Upgrading buildkite/cli/bk v1.0.0 -> v1.1.0
==> Downloading $STABLE_URL
Warning: Transient problem: timeout Will retry in 1 seconds. 3 retries left.
Warning: Transient problem: timeout Will retry in 2 seconds. 2 retries left.
Warning: Transient problem: timeout Will retry in 4 seconds. 1 retries left.
##O=#  #
curl: (6) Could not resolve host: $STABLE_URL
Error: Failed to download resource "bk"
Download failed: $STABLE_URL
```

Looks like https://github.com/buildkite/cli/blob/master/.github/tap-release.yml#L11 `$STABLE_URL` and `$STABLE_SHA256` are not being replaced with a value; but `$STABLE_VERSION` is correctly being given a value.

My local `bk` formula is:

```
# Generated by https://github.com/buildkite/cli/blob/master/.github/tap-release.yml

class Bk < Formula
  homepage "https://github.com/buildkite/cli"
  desc "A command line interface for Buildkite."
  url "$STABLE_URL"
  version "v1.1.0"
  sha256 "$STABLE_SHA256"

  def install
    mv "bk-darwin-amd64-1.1.0", "bk"
    bin.install "bk"
  end

  test do
    assert_match "bk #{version}", shell_output("#{bin}/bk --version 2>&1")
  end
end
```

samattyro:
Is there any update for fixing or workaround please.

orien:
The examples in the [`tap-release`](https://github.com/toolmantim/tap-release) readme specify slightly different variable names.

```diff
-    url      "$STABLE_URL"
-    sha256   "$STABLE_SHA256"
+    url      "$STABLE_ASSET_URL"
+    sha256   "$STABLE_ASSET_SHA256"
```

Could this be the fix?

samattyro:
Thank orien.  I manually updated my bk.rb file to "$STABLE_ASSET_URL" and "$STABLE_ASSET_SHA256" with no luck.  I feel they should be replaced with values as the variable 'version' does.

orien:
@samattyro You should be able to manually update your `bk.rb` file with the following values as a work around:

```ruby
url "https://github.com/buildkite/cli/releases/download/v1.1.0/bk-darwin-amd64-1.1.0"
[...]

PR #85, opened 2020-05-21
Update the url and version for tap formula

JuanitoFatas:
Thanks @orien for the tips in #83.


timc13:
any chance this will get merged soon? =) this is blocking us from installing the `cli` and getting started on developing pipelines

toolmantim:
Sorry about that @timc13! That should be all fixed now.

PR #93, opened 2020-09-02
Fix usage example in README

rohansingh:
There is no `bk pipelines`, it should be `bk pipeline`.


rohansingh:
(ping @pda)

pda:
Thanks for the fix, and the reminder ping! I totally missed it the first time. 🙏🏼

PR #100, opened 2020-12-01
Convert GitHub auth process from "web application flow" to "device flow"

yob:
When this app was originally written, the web application flow was the only way to collect a GitHub oauth token for the user.

However, [in July 2020 GitHub added support for the device flow](https://github.blog/changelog/2020-07-27-oauth-2-0-device-authorization-flow/). The device flow is nicer because only the oauth client ID needs to be distributed in the binary - there's no need for an oauth secret any more.

[GitHub docs for the device flow are here](https://docs.github.com/en/free-pro-team@latest/developers/apps/authorizing-oauth-apps#device-flow).

The first-run configure process now looks like this. Note the 8-digit code, which the user enters into their browser in an authenticated github.com session.

```
$ go run ./cmd/bk configure
──── Ok! Let's get started with configuring bk 🚀


──── Configuring Buildkite GraphQL credentials

Create a GraphQL token at https://buildkite.com/user/api-access-tokens/new. Make sure to tick the GraphQL scope at the bottom.

GraphQL Token: user@example.com ✅

Enter passphrase to unlock /home/jh/.buildkite/keyring/: 
Securely stored GraphQL token! 💪
Wrote configuration to /home/jh/.buildkite/config.json 📝

For now, we will assume you are using GitHub. Support for others is coming soon! 😓

──── Let's configure your github.com credentials 💻

We need to authorize this app to access your repositories. This authorization is stored securely locally, buildkite.com never gets access to it.

In a moment, we'll print a unique code and open a github.com URL in your default browser. To authenticate bk, enter the unique code into the browser.

Press enter to continue


User Code: XXXX-YYYY
....

Authenticated as yob ✅
Securely stored GitHub token! 💪

Ok, you are good to go!
```


github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
-)
-
-var (
-	oauthConf = oauth2.Config{
-		ClientID:     "d5a2938d576279fbc995",
-		ClientSecret: "aadcab62248629c3d0018b9e54e4d3fc37fdd29e",
-		Scopes:       []string{"user:email", "repo"},
-		Endpoint:     githuboauth.Endpoint,
-	}
+	oauthClientId       = "d5a2938d576279fbc995"
```
pda, line 20: I can't think of any material benefit, but is it worth pulling this up to a const or something, rather than inlining it?

github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
-	handler := githubAuthHandler{
-		OAuthStateString: "da39a3ee5e6b4b0d3255bfef95601890afd80709",
-		TokenCh:          make(chan *oauth2.Token),
+	if resp.StatusCode != 200 {
+		return nil, errors.New(fmt.Sprintf("Github returned non-200 response (%d)", resp.Status))
```
pda, line 45:
```suggestion
		return nil, fmt.Errorf("Github returned non-200 response (%d)", resp.Status)
```

github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
+
+		resp, err = http.Post(fmt.Sprintf("%s?client_id=%s&device_code=%s&grant_type=%s", githubStageThreeUri, oauthClientId, deviceCode, githubGrantType), "application/json", nil)
+		if err != nil {
+			return nil, err
 		}
-		token, err := oauthConf.Exchange(oauth2.NoContext, req.FormValue("code"))
+		defer resp.Body.Close()
+
+		if resp.StatusCode != 200 {
+			return nil, errors.New(fmt.Sprintf("Github returned non-200 response (%d)", resp.Status))
```
pda, line 101:
```suggestion
			return nil, fmt.Errorf("Github returned non-200 response (%d)", resp.Status)
```

github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
+	params, err := url.ParseQuery(string(body))
+	if err != nil {
 		return nil, err
 	}
 
-	select {
-	case token := <-handler.TokenCh:
-		return token, ln.Close()
-	case err := <-handler.ErrCh:
+	deviceCode := params["device_code"][0]
```
pda, line 56: This (and other similar lines) probably needs some guards / bounds-checking to prevent runtime panic, maybe?

github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
+		return nil, err
+	}
+
+	if err := open.Run(verificationUri); err != nil {
+		return nil, err
+	}
+
+	for {
+		if time.Now().After(expiresAt) {
+			continue
```
pda, line 90: Do you want `break` here? Or am I misremembering how `for`/`continue` works.

github/oauth_server.go
```diff
@@ -13,82 +17,107 @@ import (
...
+	intervalInt, err := strconv.Atoi(interval)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := open.Run(verificationUri); err != nil {
+		return nil, err
+	}
+
+	for {
```
pda, line 88: There's probably a fancy Go way using a ticker and a timer, and `select` on the two channels.
But this is also fine :)

pda (commented):
Looking good!

github/oauth_server.go
yob: So something like this?

```diff
diff --git a/github/oauth_server.go b/github/oauth_server.go
index 73c3130..f8f172e 100644
--- a/github/oauth_server.go
+++ b/github/oauth_server.go
@@ -16,7 +16,7 @@
        "golang.org/x/oauth2"
 )
 
-var (
+const (
        oauthClientId       = "d5a2938d576279fbc995"
        oauthScopes         = "user:email,repo"
        githubStageOneUri   = "https://github.com/login/device/code"
```

github/oauth_server.go
yob: I dabbled with this for a while, but then realised that github will return an error when the 15 minutes is up and the user token has expired. As a result I've turned this into an endless loop that exits on either:

1. an access token is generated
2. github returns an error, one of which can be "time is up, try again"

yob:
It's a bit depressing what this does to the line count, but overall I'm quite find of the device flow. It's easy to understand, and easy to implement.

Screenshot from 2020-12-03 00-00-10

I think I want to rewrite history and tidy up the commits, but before I do that I'm keen on feedback on the current code :pray:

pda (approved):
```
$ ./bk configure github
──── Let's configure your github.com credentials 💻

We need to authorize this app to access your repositories. This authorization is stored securely locally, buildkite.com never gets access to it.

In a moment, we'll print a unique code and open a github.com URL in your default browser. To authenticate bk, enter the unique code into the browser.

Press enter to continue


User Code: ____-____
.............

Authenticated as pda ✅
Securely stored GitHub token! 💪
```

image

jayco (approved):
LGTM 🎉

pda (approved):

yob:
Cool.

I have to duck out for a few hours, but I'll rewrite the commit history when I get back and merge.

yob:
This was well timed - the [device flow went GA](https://github.blog/changelog/2020-12-04-oauth-2-0-device-authorization-flow-now-ga/) 2 days after this merged.

PR #101, opened 2020-12-03
Update keyring & go-keychain for macOS 11 fix

pda:
Otherwise the build fails on macOS 11 Big Sur with:

```
# github.com/keybase/go-keychain
cgo-gcc-prolog:203:11:
    warning: 'SecTrustedApplicationCreateFromPath' is deprecated: 
    first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecTrustedApplication.h:59:10:
    note: 'SecTrustedApplicationCreateFromPath' has been explicitly marked deprecated here
```

Lots more background in:

- https://github.com/99designs/keyring/issues/56#issuecomment-566256653
- https://github.com/99designs/keyring/issues/66
- https://github.com/keybase/go-keychain/pull/55
- https://github.com/99designs/aws-vault/pull/427

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/buildkite/cli
 go 1.14
 
 require (
-	github.com/99designs/keyring v1.1.3
+	github.com/99designs/keyring v1.1.6
 	github.com/ahmetb/go-cursor v0.0.0-20131010032410-8136607ea412
 	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
 	github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -37,3 +37,5 @@ require (
 	google.golang.org/appengine v1.0.0 // indirect
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 )
+
+replace github.com/keybase/go-keychain => github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
```
yob, line 41: Could we also change all references from `github.com/keybase/go-keychain` to `github.com/99designs/go-keychain`, but this indicates we expect the replacement to be temporary?

yob (approved):
I don't have a go env setup on macos yet so I can't test this, but it seems fine :+1:

go.mod
pda: I think it's trickier than that because it's an indirect dependency, via 99designs/keyring; I think we'd have to vendor that and patch its imports too. I guess this is what `replace` was made for; I've never used it before, I just cargo-culted @mtibben's comment advice ;)

JuanitoFatas (approved):

ticky:
Oh you already fixed this! Haha

PR #107, opened 2021-02-28
Build linux/arm64 and darwin/arm64 binaries

### LINKED ISSUE #104: Darwin arm64 binaries
Could we please add darwin/arm64 binaries to the release process?

sj26:
Build arm64 binaries for both linux and darwin. Fixes #104.


pda:
Rebased onto #109 (Go 1.16)

Issue #114, opened 2021-11-08
Consider removing keyring integration?


lox:
Whenever I come back to work on `cli`, I find something has broken the https://github.com/99designs/keyring integration. It's laden with really unpleasant native stuff and I don't reckon it's worth it.

Thoughts on removing it and relying on a json file in a config dir with tokens in it? The github token is short-lived, if it was possible to switch the GraphQL token to something more short-lived it would be fine IMO. 

P.S. If there is a better spot to have discussions on the bk cli, lemme know.

PR #115, opened 2021-11-08
Remove keyring entirely and use config json

lox:
This removes entirely the fancy keyring integration in favour of a much simpler config file. 

Why? Every time I come back to this tool to iterate on it, a weird native interface library is broken and it won't build. Beyond that, it means this tool needs binary builds (for code-signing) which is awkward and impractical.

This would be a very breaking change! Closes https://github.com/buildkite/cli/issues/114.


dontlaugh:
I built your branch, and it worked great

sj26:
Thanks @lox. I mostly agree, and err on the side of development productivity.

How painful do you think it is to have tokens in files? Is there another way to do keychain tokens nicely? Every time I explore that space it seems pretty spicy.

This repo also doesn't have a clear caretaker at the moment. I suspect it would be the decision of whoever wants to pick this up and start iterating on it. So this might need to wait for that moment.

pda:
Yeah, seems reasonable.
I think this project needs to be as easy to occasionally-maintain as possible, so this would be a win.

I wonder if there's anything else that should be bundled into a v2.x release, given we'll probably need to bump major version for the breaking change 🤔

I guess we can merge in the mean time, to at least get this on `main`.

pda:
Rebased onto `main` to get https://github.com/buildkite/cli/pull/131 and added a small README tweak.

@moskyb and I have tested it locally, including the whole GraphQL token and GitHub OAuth flow, verifying the keys end up in the config file, and can be used to list pipelines 👌🏼

pda (approved):

pda:
Hrmm… not sure if this is relevant to this PR, but I thought the removal of keyring would ideally make this not happen when running `bk --help`:

image

pda:
I think this is just macOS these days… it's a downloaded binary, so it “warns” the user.
`xattr -d com.apple.quarantine <filename>` did the trick.

PR #124, opened 2022-05-16
Prefer /usr/bin/env bash

wpcarro:
Instead of `/bin/bash`, which may help this script support slightly more
systems (e.g. NixOS), where `/bin/bash` doesn't exist.


sj26:
This script is only run by CI and shouldn't be used within the cli utility. Does this cause issues for you?

pda (approved):
Seems reasonable 👍🏼
Good job working around it over in https://github.com/NixOS/nixpkgs/pull/173311 😅

PR #125, opened 2022-05-17
Make BUILDKITE_PIPELINE_DEFAULT_BRANCH configurable

wpcarro:
Because `BUILDKITE_PIPELINE_DEFAULT_BRANCH` is *not configurable* through an
environment variable, I think a special purpose env var for `buildkite-cli` may
make sense. As such, I defined the `BUILDKITE_CLI_` prefix for this case. I'm
open to alternatives.

Also defaulting to `main` instead of `master`, since this is the new GH default.

### DIFF
--- a/local/server.go
+++ b/local/server.go
@@ -387,7 +387,7 @@ func (a *apiServer) handleAcceptJob(w http.ResponseWriter, r *http.Request, jobI
 		`BUILDKITE_AGENT_META_DATA_QUEUE`:     `default`,
 		`BUILDKITE_REBUILT_FROM_BUILD_ID`:     ``,
 		`BUILDKITE_REBUILT_FROM_BUILD_NUMBER`: ``,
-		`BUILDKITE_PIPELINE_DEFAULT_BRANCH`:   `master`,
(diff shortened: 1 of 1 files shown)


local/server.go
```diff
@@ -387,7 +387,7 @@ func (a *apiServer) handleAcceptJob(w http.ResponseWriter, r *http.Request, jobI
 		`BUILDKITE_AGENT_META_DATA_QUEUE`:     `default`,
 		`BUILDKITE_REBUILT_FROM_BUILD_ID`:     ``,
 		`BUILDKITE_REBUILT_FROM_BUILD_NUMBER`: ``,
-		`BUILDKITE_PIPELINE_DEFAULT_BRANCH`:   `master`,
+		`BUILDKITE_PIPELINE_DEFAULT_BRANCH`:   getenv("BUILDKITE_CLI_PIPELINE_DEFAULT_BRANCH", "main"),
```
sj26, line 390: Nice idea!

I think given the special context of bulidkite-cli it makes sense to inherit this value from the outer environment. So instead of a new env, perhaps we could use the same env:

```suggestion
		`BUILDKITE_PIPELINE_DEFAULT_BRANCH`:   getenv("BUILDKITE_PIPELINE_DEFAULT_BRANCH", "main"),
```

moskyb (approved):

PR #135, opened 2022-09-21
Add fallback to XDG dirs

mmlb:
This PR adds the appropriate xdg base dir as the default/fallback for both config file and emoji cache paths. This uses adrg/xdg which has seemingly good choices for non-linux platforms. Either way though, its only a fallback so if a pre-existing directory is found its used. This will make buildkite/cli more respectful of the user's $HOME dir instead of just dumping yet another hidden directory with mixed config&data that needs to be thought about for backups/sync/ls purposes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/adrg/xdg"
 	homedir "github.com/mitchellh/go-homedir"
 	"golang.org/x/oauth2"
 )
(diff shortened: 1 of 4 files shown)


mmlb:
quick question, should I print a log message to stderr if the old directory is found?

config/config.go
```diff
@@ -21,17 +22,24 @@ type Config struct {
...
 	file := os.Getenv("BUILDKITE_CLI_CONFIG_FILE")
-	if file == "" {
-		home, err := homedir.Dir()
-		if err != nil {
-			return "", err
-		}
+	if file != "" {
+		return file, nil
+
+	}
```
triarius, line 31: Probably best to limit the scope of `file` with
```
if file := os.Getenv("BUILDKITE_CLI_CONFIG_FILE"); file != "" {
	return file, nil
}
```

triarius (approved):
Thanks for this @mmlb. I think it's fine to not print to stderr when the old directory is not found.

I had a small nitpick, but since we took a long time to get back to you, I'm happy to merge this right away.

config/config.go
mmlb: Makes sense, want me to follow up with a PR or will you take care of it?

triarius: Feel free. We've got a bit on our plates and don't have time to prioritise something like this at the moment.

PR #137, opened 2023-01-13
Update README.md to add Apple M-series chip install command

logankilpatrick:
Ran into this issue trying to install, I always forget you need to do this on the new M-series macs so adding for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,12 @@ On macOS, you can install with [Homebrew](https://brew.sh):
 brew install buildkite/cli/bk
 ````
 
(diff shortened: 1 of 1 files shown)


jacobjohnsonutra:
Begin

jacobjohnsonutra (approved):

sj26:
Thanks for the contribution! But this tool should work on Apple silicon without specifying the arch now — that's how I run it — so these instructions should not be required. Please try it again and re-open if you're still having trouble 🙏🏻

PR #139, opened 2023-02-02
Enable async pipeline upload flow

triarius:
The API and agent will soon support a pipeline upload where retries will not require re-uploading the same pipeline.
However, the agent CI uses the buildkite cli, and was failing for that PR because the cli did not handle this new async pipeline upload method.

This PR adds the routes that the agent would call and responds in the appropriate way to allow the agent CI job to pass.

Here is the agent CI job that uses the bk cli passing while using the async pipeline upload:
2023-02-02T23:26:55,935276163+11:00


sj26 (approved):
Looks good

PR #140, opened 2023-02-04
Add XDG path for artifacts

triarius:
Fallback [XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) paths for the config and the emoji cache were introduced in https://github.com/buildkite/cli/pull/135. This extends this to the artifact storage path. So if the path `.buildkite/local` does not exist, artifacts will be stored in `$XDG_DATA_HOME/buildkite/artifacts` instead. This will fall back to `~/.local/share/buildkite/artifacts`.

There is also some code cleanup, and the removal of "ignored" from the emoji cache path.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,8 +35,9 @@ func Path() (string, error) {
 		}
 	}
 
-	// xdg.CacheFile will create buildkite dir if it doesn't exist but does not touch/create config.json
-	return xdg.ConfigFile("buildkite/config.json")
+	// xdg.ConfigFile will create buildkite dir if it doesn't exist but does not touch/create config.json
+	return xdg.ConfigFile(filepath.Join("buildkite", "config.json"))
+}
 }
 
 func EmojiCachePath() (string, error) {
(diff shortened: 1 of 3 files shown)


local/server.go
```diff
@@ -723,7 +723,7 @@ func (a *apiServer) handleArtifactsUpload(w http.ResponseWriter, r *http.Request
 	}
 	filename := matches[1]
 
-	cacheDir, err := artifactCachePath()
+	cacheDir, err := config.ArifactStoragePath()
```
moskyb, line 726:
```suggestion
	cacheDir, err := config.ArtifactStoragePath()
```

config/config.go
```diff
@@ -24,22 +24,52 @@ type Config struct {
...
 		}
 	}
 
-	// xdg.CacheFile will create buildkite dir if it doesn't exist but does not touch/create config.json
-	return xdg.ConfigFile("buildkite/config.json")
+	// xdg.ConfigFile will create buildkite dir if it doesn't exist but does not touch/create config.json
+	return xdg.ConfigFile(filepath.Join("buildkite", "config.json"))
+}
+
+func ArifactStoragePath() (string, error) {
```
moskyb, line 42:
```suggestion
func ArtifactStoragePath() (string, error) {
```

local/server.go
moskyb, line 726:
```suggestion
	cacheDir, err := config.ArtifactStoragePath()
```

moskyb (approved):

triarius:
I forced pushed to fix a typo in the commit message of a commit that fixes a typo in the code.

PR #141, opened 2023-02-04
Bump golang to v1.20

triarius:
Upgrading to golang 1.20 was not straightforward, as there was a change in how the golang standard library's `"net/http"` package reports an uploaded file's filename in a request. Previously `header.Filename` returned the absolute filename, but now it returns the base of the filename. This is due to a security fix introduced in go 1.17: https://go.dev/doc/go1.17#mime/multipart

Unfortunately, the Buildkite API preserved most of the filename of that was uploaded using its absolute filename, so we have to preserve this behaviour. Fortunately, I manged to recover the absolute filename from the `Content-Disposition` header.

This was not picked up in the tests, as there are no absolute artifact uploads in the test pipeline. This is why the initial commit to bump the versions, [8ad27de](https://github.com/buildkite/cli/pull/141/commits/8ad27de2717202d27a6d6c49b95e547bbad4f512) builds [successfully](https://buildkite.com/buildkite/buildkite-cli/builds/90). However, introducing an absolute upload in [592f4de](https://github.com/buildkite/cli/pull/141/commits/592f4de11be65335ad9977ee277a2dc4564238ef) breaks the [build](https://buildkite.com/buildkite/buildkite-cli/builds/101). This is fixed in the final commit in the manner mentioned in the previous paragraph.

### DIFF
--- a/local/server.go
+++ b/local/server.go
@@ -716,7 +716,12 @@ func (a *apiServer) handleArtifactsUpload(w http.ResponseWriter, r *http.Request
 	defer file.Close()
 
 	contentDisposition := header.Header.Get("Content-Disposition")
-	filename := filenameRegexp.FindStringSubmatch(contentDisposition)[1]
+	matches := filenameRegexp.FindStringSubmatch(contentDisposition)
+	if len(matches) < 2 {
+		http.Error(w, "filename missing from Content-Disposition header", http.StatusBadRequest)
+		return
+	}
+	filename := matches[1]
 
 	cacheDir, err := artifactCachePath()
 	if err != nil {
--- a/local/server.go
+++ b/local/server.go
@@ -74,8 +74,8 @@ type jobEnvelope struct {
 }
 
 var (
-	uuidRegexp = regexp.MustCompile(
-		"([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})")
+	uuidRegexp     = regexp.MustCompile("([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})")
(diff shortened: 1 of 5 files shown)


local/server.go
```diff
@@ -715,14 +715,16 @@ func (a *apiServer) handleArtifactsUpload(w http.ResponseWriter, r *http.Request
 	}
 	defer file.Close()
 
+	contentDisposition := header.Header.Get("Content-Disposition")
+	filename := filenameRegexp.FindStringSubmatch(contentDisposition)[1]
```
DrJosh9000, line 719: What if Content-Disposition doesn't contain `filename=".."`, for whatever reason?

triarius: Good point. Looks like it's not required by [some](https://www.rfc-editor.org/rfc/rfc1867#section-3.3) [RFCs](https://www.rfc-editor.org/rfc/rfc7578#section-4.2).
It's a "SHOULD" not a "MUST" in the later one.

However, the agent does seem to always provide it. It's a little convoluted why and how, but I think it ultimately because of this method: https://go.dev/src/mime/multipart/writer.go#L140. Obviously, this may change if the implementation of `CreateFormFile` in the standard library changes, this may seem unlikely, but such a change is why I'm writing this code in the first place 🤣 If that happens, it's likely that the [integration test](https://buildkite.com/buildkite/agent/builds/5191#0186217e-33a4-4bd0-9a9f-ecf459643f1e) will fail then as well. However, we do seem to unit test the old implementation [here](https://github.com/buildkite/agent/blob/81cbb6fea466b9f3c810be8260b78389bbd5396e/agent/form_uploader_test.go#L58-L61). So I suggest we change that test to use the same implementation as here, and write a unit test that distinguishes between an absolute and relative upload as well. This would be a PR in the agent repo.

As for what to do here, I'm going to make sure it does not panic if the regex does not match, and I'll surface an error if it's not.

DrJosh9000: > I'm going to make sure it does not panic if the regex does not match, and I'll surface an error if it's not.

Sounds good to me!

DrJosh9000 (approved):
LGTM 😎

PR #142, opened 2023-02-08
Fix TestPipelineParsing flakiness

DrJosh9000:
We apparently support three different kinds of plugin configuration in a pipeline (see the command step `UnmarshalJSON` method for details). This fixes the test data to use the form with stable ordering.

### DIFF
--- a/local/pipeline_test.go
+++ b/local/pipeline_test.go
@@ -209,12 +209,12 @@ func TestPipelineParsing(t *testing.T) {
 							"echo hello world",
 							"pwd"
 						],
-						"plugins": {
-							"blah-blah/blah#v0.0.1": null,
-							"blah-blah/another#v0.0.1": {
+						"plugins": [
+							{"blah-blah/blah#v0.0.1": null},
+							{"blah-blah/another#v0.0.1": {
 								"my_config":"totes"
-							}
-						}
+							}}
+						]
 					}
 				]
 			}`,


local/pipeline_test.go
```diff
@@ -209,12 +209,12 @@ func TestPipelineParsing(t *testing.T) {
...
 						],
-						"plugins": {
-							"blah-blah/blah#v0.0.1": null,
-							"blah-blah/another#v0.0.1": {
+						"plugins": [
+							{"blah-blah/blah#v0.0.1": null},
+							{"blah-blah/another#v0.0.1": {
 								"my_config":"totes"
-							}
-						}
```
triarius, line 217: From what I recall of what you found out when switching the YAML parser on the agent, the unordered map YAML syntax is secretly ordered when the agent parses. So it is the case that JSON syntax being removed here is not something the agent would send to the API?

triarius (commented):
Wow. This one was right up your alley!

I have a question, will approve depending on that.

local/pipeline_test.go
DrJosh9000: I didn't really want to spend the afternoon fixing the pipeline parser here - it would be a lot more involved.

The previous JSON is exactly what the agent would send if the plugins were all in one map. The new JSON is what it would send if the plugins were (more correctly) in a sequence of one-item maps.

Maybe the agent should convert the first form into the second?

triarius: Ok makes sense. So the JSON parsing is not quite the same in the CLI as the API in that it won't preserve order in an unordered object 🙉 It's probably worth recording that as a ticket in the backlog, or even an issue in this repo with some context.

> Maybe the agent should convert the first form into the second?

I think that's prudent, but might not be necessary. I suspect it "just works" in the API because something something insertion order preserving iteration for hashes in ruby.

It would be good to make this explicit in all 3 of the API, CLI and agent though.

triarius (approved):

local/pipeline_test.go
DrJosh9000: > It would be good to make this explicit in all 3 of the API, CLI and agent though.

Yes. And if it were up to me, I'd make it so we never accept a JSON object or YAML map where we care about the order.

triarius: We can only dream...for now

PR #149, opened 2023-08-24
Use a configure command to set auth

mcncl:
In the current `bk` we use `configure` to set the access token values. It might be nice to keep it that way.

Also added some error handling to the `Read`


cmd/bk/main.go
```diff
@@ -18,13 +18,18 @@ func main() {
...
+	v.SetConfigFile(config.ConfigFile())
+	v.AutomaticEnv()
 	// attempt to read in config file but it might not exist
-	_ = viper.ReadInConfig()
+	if err := v.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			fmt.Print(fmt.Errorf("Unable to find config file at %s: %s", config.ConfigFile(), err))
+			return 1
+		}
+	}
```
jradtilbrook, line 30: I thought about this handling but that means the config file has to exist first which makes the configure command kinda redundant. We should be able to operate the entire bk cli without it

pkg/cmd/configure/configure.go
```diff
@@ -9,13 +9,13 @@ import (
...
-func NewCmdInit(viper *viper.Viper) *cobra.Command {
+func NewCmdConfigure(v *viper.Viper) *cobra.Command {
 	var force bool
 
 	cmd := &cobra.Command{
-		Use:   "init",
+		Use:   "configure",
 		Args:  cobra.ExactArgs(0),
-		Short: "Initialise authentication token",
+		Short: "Configure Buildkite authentication token",
```
jradtilbrook, line 18: Maybe even 
```suggestion
		Short: "Configure Buildkite API token",
```

jradtilbrook (changes requested):

cmd/bk/main.go
mcncl: It also handles the fact that between `configure` and `<command>` the user has deleted their config file.

jradtilbrook: I've checked this out locally because it wasnt making sense. If the file doesn't exist it should error out here and make the CLI unusable. But it wasn't doing that.

Turns out that's because the err type is actually `*fs.PathError` so it never gets into the if condition to exit early.

But we don't want that anyway. We need to be able to invoke the command in order to create the config file. So the config file not existing is valid

jradtilbrook (approved):

PR #150, opened 2023-08-24
Init command with checks on existing file.

mcncl:
`init` will check to see if a pipeline.yaml file exists in the `.buildkite` directory of the current project. If it does then it won't initialise another one, if it doesn't then it will create the `.buildkite` dir and a stubbed `pipeline.yaml` file.

Happy for any changes to be made to this that are desired.


jradtilbrook:
I like it. I think we should check for all files that the agent supports though https://buildkite.com/docs/agent/v3/cli-pipeline#uploading-pipelines-description but I can take it from here 👍

jradtilbrook (approved):

PR #156, opened 2023-09-13
Add LIST command to agent

mcncl:
## Changes
This implements a `list` command on the `agent` that lists all agents within a users organisation.

I've also added support for setting the `--output|-o` to either JSON or YAML. See below.

bk_list_1

bk_list_2

### Update
I've removed the table printing for now; the sensible default feels like JSON as folks can actually do somethting with the returned data.

We can add the table back in if we want to, but other than a visual thing it serves no purpose and would actually be more difficult to use; with JSON we could use `jq` to use the returned data, for example.

CleanShot 2023-09-18 at 09 51 44

### Additional Update
* I've refactored the `print` functions to accept anything so they can be re-used
* I've created a single `view` command which takes in an `--id` and then produces output of the given agent
* Commands, where applicable, now all accept an `--output|-o` flag to set either `json` or `yaml`. Default is currently `json`

CleanShot 2023-09-18 at 17 17 09


pkg/cmd/agent/list.go
```diff
@@ -0,0 +1,35 @@
...
+		DisableFlagsInUseLine: true,
+		Use:                   "list",
+		Args:                  cobra.NoArgs,
+		Short:                 "Lists the agents for the current organization",
+		Long: heredoc.Doc(`
+            Command to list all agents for the current organization.
+
+
+            Only running agents are listed by default. To list all agents, use the
+            --all flag.
```
jradtilbrook, line 22: Remove this since there is no flag?

pkg/cmd/agent/list.go
```diff
@@ -0,0 +1,35 @@
...
+		`),
+		RunE: func(cmd *cobra.Command, args []string) error {
+            agents, _, err := f.RestAPIClient.Agents.List(f.Config.Organization, &buildkite.AgentListOptions{})
+
+            for i, agent := range agents {
+                fmt.Printf("%d: %s. ID: %s. State: %s\n", i+1, *agent.Name, *agent.ID, *agent.ConnectedState)
```
jradtilbrook, line 28: I think there is a table output format but this will do for now 👍

jradtilbrook (changes requested):

internal/printer/printer.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type Output string
+
+const (
+	JSON Output = "json"
+	YAML Output = "yaml"
+)
+
+func PrintOutput(output Output, p any) (string, error) {
```
jradtilbrook, line 16: In terms of code maintenance, I'd like to see this abstracted more and configured as part of startup. So something early on from the root command could read the output flag and set a global config somewhere. That way its not repeated in individual commands and those commands just call a basic "print" type function that knows what format. 

But another thing to refactor for now

pkg/cmd/agent/agent.go
```diff
@@ -6,11 +6,18 @@ import (
...
 	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	"github.com/buildkite/go-buildkite/v3/buildkite"
 	"github.com/spf13/cobra"
 )
 
+type Agent buildkite.Agent
+
+func (a Agent) Print(output printer.Output) {
+}
+
```
jradtilbrook, line 20: Was this the beginning of something you've abandoned now?

pkg/cmd/agent/view.go
```diff
@@ -0,0 +1,48 @@
...
+			data, err := printer.PrintOutput(printer.Output(output), agent)
+			if err != nil {
+				return err
+			}
+			fmt.Println(data)
+			return nil
+		},
+	}
+
+	cmd.Flags().String("id", "", "The ID of the agent")
```
jradtilbrook, line 45: I don't think this should be a flag, it should be an arg? eg [`gh issue view <num>`](https://primer.style/cli/foundations/language)

pkg/cmd/validation/uuid.go
```diff
@@ -0,0 +1,12 @@
+package validation
+
+import "github.com/google/uuid"
+
+// UUID validates that the given string is a valid UUID.
+func ValidateUUID(s string) error {
```
jradtilbrook, line 6: I'm not sure the value in this, either we'll get an error from the API or can validate it early.

But I think it is useful for users to be able to use the agent URL as well and we parse that out. So they can easily copy/paste from the web

jradtilbrook (changes requested):
Looking really good

I like how we have json and yaml options now 👌  But I disagree on json being the default. I think defaulting to a table/more human readable/consumable format is better and then if users are going to make use of this in some automation, they will add the -o json flag. But we can follow up with that later.

I've left a few comments here but most we can follow up with in another PR. The only one I'd like to see changed now is the `--id` flag become an argument instead

pkg/cmd/agent/view.go
mcncl: Depends entirely on which CLI you're trying to replicate; in `aws` it's a flag.

pkg/cmd/validation/uuid.go
mcncl: Hmm, my personal idea with the CLI is that it reduces/entirely removes the requirement to go to the web. I don't really see much value in a CLI in which you need to go to the web to get the values.

mcncl:
@jradtilbrook the table output feels like a preference over a function? Our API outputs JSON when a REST call is made, other tools such as `aws cli` output JSON when you `describe` a _thing_. A table is "nice" if the only output is the agent name and ID, but should data be masked if they opt to not select JSON or YAML?

sj26:
Nice!

Instead of `view` may I suggest `get` to mirror the command structure from the agent:

https://github.com/buildkite/agent/blob/main/clicommand/meta_data_get.go

mcncl:
@sj26 that feels like a fair suggestion and I'm all for consistency in the experience across our tooling. I've made that change.

Thank you!

sj26:
(Love to see the CLI getting some love!)

jradtilbrook:
This has been implemented although it doesnt have the output formatting which we will come back to later

PR #159, opened 2023-09-27
SUP-1420: README addition

james2791:
README addition (from `main` with small tweaks). 

Installation heading and instructions to be added additionally once the details for `3.x` are known. (Same applies with `usage`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,20 @@
+# bk - The Buildkite CLI
+
+[![Latest release](https://img.shields.io/github/release/buildkite/cli.svg)](https://github.com/buildkite/cli/releases/latest)
+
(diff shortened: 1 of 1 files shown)


jradtilbrook (dismissed):
That will do for now 👍

mcncl (changes requested):
Should we highlight that this `default` branch is under development? Or point to the more detailed README of the current stable version?

mcncl (approved):

james2791:
I +1 on @mcncl's addendum suggestion - added a ref to `main` and releases page for the stable releases to date

jradtilbrook (approved):

PR #166, opened 2023-11-07
Add missing local run env vars

dannymidnight:
The commonly used [`buildkite-docker-plugin`](https://github.com/buildkite-plugins/docker-buildkite-plugin) [references](https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/master/commands/run.sh#L442-L453) some additional Buildkite env vars that aren't currently exposed when running a local build.

This PR adds the following env vars (and re-orders them to make it easier to add new ones)

* `BUILDKITE_PIPELINE_NAME`
* `BUILDKITE_STEP_KEY`


james2791 (approved):
LGTM - also noting this is on `main` too!

PR #171, opened 2023-12-07
SUP-1566 Augment agent stop with loading spinner

jradtilbrook:
This adds a new module to encapsulate a loading spinner using bubble tea. Doing it through bubble tea is quite complex for just a spinner while another function is running, but this should simplify the process into a neat package.

We can iterate on this design as well as we make more use of things that do some long API requests, etc.

I've updated the `bk agent stop` command to make use of this so you can see an example.  
Let me know what you think of the API


internal/io/pending_command.go
```diff
@@ -0,0 +1,78 @@
...
+		p.output = string(msg)
+		return p, tea.Quit
+	case error:
+		p.quitting = true
+		p.output = "Error: " + msg.Error()
+		return p, tea.Quit
+	}
+	var cmd tea.Cmd
+	p.spinner, cmd = p.spinner.Update(msg)
+	return p, cmd
```
jradtilbrook, line 56: Perhaps this should just be a `default:` case above? And remove the `spinner.TickMsg` case to consolidate them. Then as we make more use of this we can add more cases

PR #172, opened 2023-12-07
SUP-1566 Add agent view command

jradtilbrook:
This adds `bk agent view` command to view information about an agent.

It doesn't show everything like the web does (it doesn't seem to be available in the API). It looks like more is available in GraphQL but this should make a good enough start to continue with at least.

Example output: 
```
Jarryds-MacBook-Pro.local-1
 ID           018c4734-5fde-4cc1-aef6-05b3519e258c
 State        connected
 Version      3.58.0
 Hostname     Jarryds-MacBook-Pro.local
 User Agent   buildkite-agent/3.58.0.x (darwin; arm64)
 IP Address   125.253.50.1
 Connected    Fri, 08 Dec 2023 02:14:23 +0000
 Metadata     queue=default
```


james2791 (dismissed):
🚀  - works well!

internal/io/pending_command.go
```diff
@@ -59,7 +59,7 @@ func (p Pending) View() string {
 	if !p.quitting {
 		return fmt.Sprintf("%s %s", p.spinner.View(), p.output)
 	}
-	return p.output + "\n"
```
jradtilbrook, line 62: I don't understand why at the moment but by not printing a newline at the end, upon the final output before the program exits, the last line is swallowed. I think it might be best to check if the last character in `output` is a newline and add one if not

james2791 (approved):

PR #175, opened 2023-12-15
SUP-1567 Add bk agent list

jradtilbrook:
Add `bk agent list` to print a table of all connected agents.

This will serve as the basis for more functionality in the list output

Example output:

```
 ID                                    Name                         Status     Tags
 018c6dcf-917d-4e0a-98d1-20e5f9f35bb2  Jarryds-MacBook-Pro.local-1  connected  key1=val1, key2=val2, queue=default
```


pkg/cmd/agent/list.go
```diff
@@ -0,0 +1,77 @@
...
+					}
+					tags := strings.Join(agent.Metadata, ", ")
+					if max := len(tags); max > maxTags {
+						maxTags = max
+					}
+					row := []string{
+						*agent.ID, *agent.Name, *agent.ConnectedState, tags,
+					}
+
+					rows = append(rows, row)
```
mcncl, line 55: An alternative solution to reduce the repetition might be something like:

```go
var rows []table.Row
maxLengths := map[string]int{"ID": 0, "Name": 0, "Status": 0, "Tags": 0}

for _, agent := range agents {
    agentData := map[string]string{
        "ID":     *agent.ID,
        "Name":   *agent.Name,
        "Status": *agent.ConnectedState,
        "Tags":   strings.Join(agent.Metadata, ", "),
    }

    // Update maximum lengths
    for key, value := range agentData {
        if len(value) > maxLengths[key] {
            maxLengths[key] = len(value)
        }
    }

    rows = append(rows, []string{agentData["ID"], agentData["Name"], agentData["Status"], agentData["Tags"]})
}
```

No real preference, I just initially didn't grasp what the `maxX` were until I saw the `{Title: "ID", Width: maxID}` lines below.

jradtilbrook: Thats a bit nicer. I was just shoehorning this in for now since I'd like to reimplement the list output completely in some future tickets. This sort of thing should just be handled for us with that. Still looking into what that would look like though

mcncl (dismissed):

james2791 (approved):
Noting above too (for eventuality ☝️ )

james2791 (approved):

PR #181, opened 2023-12-27
SUP-1652: Filterable `bk agent list` via REST API query parameters

james2791:
This change adds three optional flags (`name`, `hostname` and `version`) that are passed through to the REST API client in the CLI as `AgentListOptions` (query parameters) when performing the `List` function to obtain a list of respective agents server-side. 

I've also added the hostname/version to the outputted table for easier viewing of the agents that matched the filtering, and adjusted the column width calculation for the table for easier maintaining.


mcncl (approved):
Looks good, I generally think we can split this function out as it's getting quite verbose, but that can be done in the new year.

PR #182, opened 2023-12-28
Resolve project pipeline

mcncl:
In [SUP-1452](https://linear.app/buildkite/issue/SUP-1452/resolve-pipeline-from-working-dir) we discuss resolving the pipeline from either the current `dir` or allowing the user to configure it.

- [x] Check if the `buildkite.yaml` config file exists in the project root
- [x] If it exists, check that it has a `pipeline: <value>` config; if so then use that
- [x] If the config exists but doesn't define a `pipeline` then use the `dir` for that value
- [x] If the file contains anything other than `pipeline` then append the `pipeline` value as the `dir`

### DIFF
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -1,0 +1,1 @@
+pipeline: cli
--- a/buildkite.yaml
+++ b/buildkite.yaml
@@ -1,1 +1,0 @@
-pipeline: cli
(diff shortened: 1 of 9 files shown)


james2791 (approved):
🚄 🚤

PR #184, opened 2024-01-10
SUP-1682 Parallelise bulk agent stopping

jradtilbrook:
Still to do:

- [ ] ~~test BulkAgent~~ this doesnt seem very useful, its such a small/simple component
- [x] what to do if >1 agent errors. should the exit code be non-zero
- [x] add request batch (maybe using https://pkg.go.dev/golang.org/x/sync/semaphore)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,11 +12,14 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/viper v1.18.2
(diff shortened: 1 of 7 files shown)


james2791:
Works quite nicely - think this is a nice approach for other potential bulk functions too 🙂

lizrabuya (approved):
🚀

james2791 (approved):
Those semaphore additions look great - potentially something that can carry over to other commands with parallelised interaction too 👍

PR #185, opened 2024-01-12
SUP-1691:  `bk agent list` interactivity

james2791:
Added a default Bubble Tea list component for `bk agent list` - which shows the agent list output in a filterable, navigable list that has default actions for sorting, clearing, exiting etc.

The model for the list component includes the list model itself, the list of Buildkite agents from Go-Buildkite, and a quitting `bool` (the latter is used for clearing the user's stdout with a `ctrl+c` when viewing the list). Pressing `esc` while filtering will clear it and return to the results given from the API.

Each row of the list (for now) includes the agents name (primary), with the secondary row showing the UUID, version and connectivity state

### DIFF
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -71,7 +71,7 @@ func (m agentListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch msg.String() {
-		case "ctrl+c", "esc":
+		case "ctrl+c":
 			m.quitting = true
 			return m, tea.Quit
 		}
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -26,15 +26,15 @@ type agentListModel struct {
 	quitting  bool
 }
 
-func ObtainAgents(f *factory.Factory, args ...string) (*agentListModel, error) {
+func ObtainAgents(f *factory.Factory, name, version, hostname string) (*agentListModel, error) {
 	var alo buildkite.AgentListOptions
 	var items []list.Item
 
-	if args[0] != "" || args[1] != "" || args[2] != "" {
+	if name != "" || version != "" || hostname != "" {
 		alo = buildkite.AgentListOptions{
-			Name:     args[0],
-			Version:  args[1],
-			Hostname: args[2],
+			Name:     name,
+			Version:  version,
+			Hostname: hostname,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


internal/agent/list.go
```diff
@@ -0,0 +1,94 @@
...
+func (ali agentListItem) Description() string { return ali.metadata }
+func (ali agentListItem) FilterValue() string { return ali.title }
+
+type agentListModel struct {
+	agentList list.Model
+	agents    []buildkite.Agent
+	quitting  bool
+}
+
+func ObtainAgents(f *factory.Factory, args ...string) (*agentListModel, error) {
```
lizrabuya, line 29: I am not sure if it is safe to use variadic functions when we don't expect the arguments to be variable.

internal/agent/list.go
```diff
@@ -0,0 +1,94 @@
...
+	var alo buildkite.AgentListOptions
+	var items []list.Item
+
+	if args[0] != "" || args[1] != "" || args[2] != "" {
```
lizrabuya, line 33: This line has a high chance to get runtime error if ObtainAgent is called with args < 3

lizrabuya (changes requested):
Some comments on the use of variadic funcs

internal/agent/list.go
james2791: Was on the fence with it - this varidic function implantation should allow for 0:N arguments to be passed in (the 3 flag values are defaulted to empty strings to cater for the use!)

internal/agent/list.go
james2791: As above - though happy to roll to specified, typed args 🙂

internal/agent/list.go
lizrabuya: Another thing, the args are not interchangeable in the implementation but the use of variadic func make it seem like they are.

jradtilbrook: Yeah I think named args are a safer approach. we can always add to later if we need more

james2791: Switched back 👍

internal/agent/list.go
james2791: I switched these to non-variadic args

jradtilbrook (approved):
I think this is looking good. I didn't run it since our pairing but it was working well then and is the foundation for more work we can split out into parallel I think

lizrabuya (approved):
LGTM! 🚀

PR #186, opened 2024-01-16
SUP-1699 Change styling of `bk agent list`

jradtilbrook:
Screenshot 2024-01-16 at 14 07 48

### DIFF
--- a/internal/agent/item.go
+++ b/internal/agent/item.go
@@ -21,5 +21,5 @@ func (ali agentListItem) QueueName() string {
 			return strings.Split(m, "=")[1]
 		}
 	}
-	return ""
+	return "default"
 }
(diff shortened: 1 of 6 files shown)


jradtilbrook:
I think this is ready now. I'd like tests covering the output side of things for the table. But I also think its likely to change with some extra PRs that will be coming so don't want to add tests just yet because they'll need refactoring

james2791 (approved):
Looks great overall :)

One thing I am thinking is as the AgentList model becomes a bit more complex, we could potentially move those helper functions from annon functions to what Charm [describes](https://github.com/charmbracelet/bubbletea/tree/master/tutorials/commands/#one-more-thing-about-commands) in writing them returning a `tea.Cmd`, which its inner annonymous func returns a `tea.Msg`

Not a major thing though!

PR #187, opened 2024-01-16
SUP-1700: Agent view output metadata improvements 

james2791:
Added some adjustments to `bk agent view` - and namely metadata parsing

The additive `parseMetadata` takes in the list returned from REST, and splits out the `queue` name while also building a multi-line string which are both fed to the respective columns in the table output (also works for the no `queue` ala `default` case as returned from the API)

I also adjusted the agent data type's name thats returned from the API as `agentData`, given the `agent` package name would keep package naming convention consistent

Connected:

image

Stopped:

image


jradtilbrook:
I just realised we should probably add a method like `func (Agent) QueueName() string` to go-buildkite to make this simpler. @james2791 could you add a linear ticket to follow up with that

lizrabuya (dismissed):
Looks great! 👍

jradtilbrook (changes requested):
Looks good. But I think making connection state and queue name formatting consistent between commands is a good idea. Currently, `bk agent list` adds colours to those fields that I think you should mimic here

jradtilbrook (approved):

PR #191, opened 2024-01-19
SUP-1658: `bk agent list`: appender, pagination and limiting

james2791:
Added pagination and limiting for `bk agent list`. This PR includes various tweaks to the existing functionality, including:

- The `loader` model functions now serves dual purpose to initially load agents, as well as append agents to the `agentList` model.  It now returns the same (renamed) `(New)AgentItemsMsg`, with the addition of the last page (set in the model from the initial `Init`). It now requires the `page` to load (`AgentListOptions`)
- `tea.KeyMsg` handled for `down` keypresses on `Update`. When the end of the current `agentList` is reached from the selected item, a further press will call the agent models `loader` with the current API page, load and then append agents to the `agentList` model and increment the page count. The list shows a status message of `Loading more agents: page X of Y` when this is happening and starts/stops the spinner.
- If there are no more agents to load from the API, the list will show a status message of `No more agents to load!`
- `per-page` integer flag for `bk agent list`: which by default is set as 30 (default REST API limits).

The below pictures use a small tweak to the agent list delegate to show a purple `>` for easier viewing of the functionality

**Initial load** (`per-page 4`)

image

**Fetching Agents** (`per-page 4`)

image

**Last page reached - agents exhausted** (`per-page 4`)

image

To decide

- [ ] Should this also be keybindable? i.e show more agents with `l` which works as another "option" with reaching the end item in `agentList`?

### DIFF
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -17,10 +17,10 @@ type AgentListModel struct {
 	agentPerPage     int
 	agentLastPage    int
 	agentsLoading    bool
-	agentLoader      func(int, int) tea.Cmd
+	agentLoader      func(int) tea.Cmd
 }
 
-func NewAgentList(loader func(int, int) tea.Cmd, page, perpage int) AgentListModel {
+func NewAgentList(loader func(int) tea.Cmd, page, perpage int) AgentListModel {
 	l := list.New(nil, NewDelegate(), 0, 0)
 	l.Title = "Buildkite Agents"
 
@@ -37,14 +37,15 @@ func (m *AgentListModel) appendAgents() tea.Cmd {
 	m.agentsLoading = true
 	// Set a status message and start the agentList's spinner
 	startSpiner := m.agentList.StartSpinner()
-	setStatus := m.agentList.NewStatusMessage(("Fetching more agents"))
+	statusMessage := fmt.Sprintf("Loading more agents: page %d of %d", m.agentCurrentPage, m.agentLastPage)
+	setStatus := m.agentList.NewStatusMessage(statusMessage)
 	// Fetch and append more agents
-	appendAgents := m.agentLoader(m.agentCurrentPage, m.agentPerPage)
+	appendAgents := m.agentLoader(m.agentCurrentPage)
 	return tea.Sequence(tea.Batch(startSpiner, setStatus), appendAgents)
 }
 
 func (m AgentListModel) Init() tea.Cmd {
-	return m.agentLoader(m.agentCurrentPage, m.agentPerPage)
+	return m.appendAgents()
 }
 
 func (m AgentListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
@@ -75,7 +76,7 @@ func (m AgentListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 		}
 	// Custom messages
-	case NewAgentItemsMsg:
+	case AgentItemsMsg:
 		// When a new page of agents is received, append them to existing agents in the list and stop the loading
 		// spinner
 		allItems := append(m.agentList.Items(), msg.ListItems()...)
@@ -83,7 +84,7 @@ func (m AgentListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.agentList.StopSpinner()
 		// If the message from the initial agent load, set the last page
 		if m.agentCurrentPage == 1 {
-			m.agentLastPage = msg.LastPage
+			m.agentLastPage = msg.lastPage
 		}
 		// Increment the models' current agent page, set agentsLoading to false
 		m.agentCurrentPage++
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -33,7 +33,7 @@ func NewAgentList(loader tea.Cmd, appender func(page int) (tea.Msg, *buildkite.R
 
 func (m *AgentListModel) appendAgents(page int) tea.Cmd {
 	// Increment to next page
-	m.agentPage++ 
+	m.agentPage++
 	// Set a status message and start the agentList's spinner
 	startSpiner := m.agentList.StartSpinner()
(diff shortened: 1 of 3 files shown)


internal/agent/list.go
```diff
@@ -12,41 +12,86 @@ import (
...
 
+func (m *AgentListModel) appendAgents() tea.Cmd {
+	// Set agentsLoading
+	m.agentsLoading = true
+	// Set a status message and start the agentList's spinner
+	startSpiner := m.agentList.StartSpinner()
+	setStatus := m.agentList.NewStatusMessage(("Fetching more agents"))
+	// Fetch and append more agents
+	appendAgents := m.agentLoader(m.agentCurrentPage, m.agentPerPage)
+	return tea.Sequence(appendAgents, tea.Batch(startSpiner, setStatus))
```
jradtilbrook, line 43: I think the arguments should be backwards here right? Sequence runs each Cmd in order until they are finished, so it won't show the spinner/status until its finished loading agents?

internal/agent/list.go
```diff
@@ -12,41 +12,86 @@ import (
...
+}
+
 func (m AgentListModel) Init() tea.Cmd {
-	return m.agentLoader
+	return m.agentLoader(m.agentCurrentPage, m.agentPerPage)
```
jradtilbrook, line 47: Should this use the appendAgents method instead now? And then perhaps the status message should just be "Loading agents", or it can check what the current page is and change the status message

internal/agent/message.go
```diff
@@ -4,11 +4,14 @@ import (
 	"github.com/charmbracelet/bubbles/list"
 )
 
-type NewAgentItemsMsg []AgentListItem
+type NewAgentItemsMsg struct {
+	Items    []AgentListItem
```
jradtilbrook, line 8: Does anything external use this field? We should make it private if not

internal/agent/list.go
```diff
@@ -12,41 +12,86 @@ import (
...
+		}
+	// Custom messages
 	case NewAgentItemsMsg:
-		// when a new page of agents is received, append them to existing agents in the list and stop the loading
+		// When a new page of agents is received, append them to existing agents in the list and stop the loading
 		// spinner
-		allItems := append(m.agentList.Items(), msg.Items()...)
+		allItems := append(m.agentList.Items(), msg.ListItems()...)
 		cmds = append(cmds, m.agentList.SetItems(allItems))
+		// Stop the loading spinner
```
jradtilbrook, line 83: I dont think we need to comment every line, especially if its just the same as reading the line its documenting

pkg/cmd/agent/list.go
```diff
@@ -23,30 +24,42 @@ func NewCmdAgentList(f *factory.Factory) *cobra.Command {
...
+						LastPage: resp.LastPage,
 					}
 				}
-				return items
 			}
 
-			model := agent.NewAgentList(loader)
+			model := agent.NewAgentList(loader, 1, perpage)
 
-			p := tea.NewProgram(model, tea.WithAltScreen())
```
jradtilbrook, line 49: We still want to use the alternate screen here

internal/agent/list.go
```diff
@@ -12,41 +12,86 @@ import (
...
+	agentList        list.Model
+	agentCurrentPage int
+	agentPerPage     int
+	agentLastPage    int
+	agentsLoading    bool
+	agentLoader      func(int, int) tea.Cmd
 }
 
-func NewAgentList(loader tea.Cmd) AgentListModel {
+func NewAgentList(loader func(int, int) tea.Cmd, page, perpage int) AgentListModel {
```
jradtilbrook, line 23: I think its a sane default to just start at the first page. So no need for that argument

jradtilbrook (changes requested):

pkg/cmd/agent/list.go
james2791: Case for me removing it on tests - have brought it back with a fresh commit

internal/agent/list.go
james2791: 👍

internal/agent/list.go
james2791: I've flipped them around, oddly was getting more consistent popup/loading with this ordering, though I think its totally YMMV

jradtilbrook: hmm could be my understanding then. perhaps thats involved with the intermittent issues ive seen in #188 🤔

internal/agent/list.go
james2791: I think thats a nice idea @jradtilbrook - maybe `Loading agents: page X of Y` with combination for using `appendAgents`?

internal/agent/message.go
james2791: Technically `loader` does - also imports said `agent` package (`v3/internal/agent`) though a constructor with private fields sounds the better option - will adjust it

internal/agent/list.go
james2791: I've tweaked the message to read `Loading more agents: page X of Y` (using the current page and last page params returned from the API)

internal/agent/message.go
james2791: And tweaked - privatised the fields and made a constructor

internal/agent/list.go
james2791: Removed the perpage argument - I think it was superfluous given that its set upon the `agent list` command run. The `page` is still needed though to pass to the `AgentListOptions` in `loader` 👍

internal/agent/list.go
james2791: image

jradtilbrook: I wonder if that is a little confusing since this presents an infinite scroll list. Not a big deal though 👍

jradtilbrook (approved):

internal/agent/list.go
james2791: Agreed - we probably can change it if it gets confusing, the actual list is paged also with the amount of items that it returns 🤔

PR #192, opened 2024-01-22
SUP-1698: Keybind for viewing agent data within `bk agent list`

james2791:
Added shared functionality for viewing an agents finer details within `bk agent list`. This PR includes various tweaks to existing `list` and `view` functionality, including:

- A shared (exported) `AgentDataTable` function that constructs a Lipgloss table and is called by both `bk agent view` (as before) and now within `bk agent list` with a `v` keybind
- A half-page viewport component for the `bk agent list` model: set initially and from every `tea.WindowSizeMsg`. Split 50-50 with the agent list, and upon terminal resizing via `setComponentSizing` (both height/width of the agent list and viewport components are set)
- `v` keybind - agent data is cleared upon navigating through to other agents with an `up`/`down` keypress
- Works with filtered data (agent data loaded from the currently selected `Item`: for which `m.VisibleItems()` is polled and thus obtained from either the filtered item list or the list model's items.

**`bk agent list` with `v` keybind** (IP/UUID redacted)

image

**Helper Keys addendum**

image

### DIFF
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -12,7 +12,8 @@ import (
 	"github.com/pkg/browser"
 )
 
-var agentListStyle = lipgloss.NewStyle().Margin(1, 2)
+var agentListStyle = lipgloss.NewStyle().Padding(1, 2)
+var viewPortStyle = agentListStyle.Copy()
 
 type AgentListModel struct {
 	agentList          list.Model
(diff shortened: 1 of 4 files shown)


jradtilbrook (approved):
Like you said in out catchup its a little buggy on resize 👍 But I think worth merging in and we can follow up with fixes.

PR #193, opened 2024-01-23
SUP-1695: Keybinding for viewing agent data in web from `bk agent list`

lizrabuya:
Add keybinding for opening agent url in web from `bk agent list` cmd


jradtilbrook (approved):

james2791 (approved):
:shipit:

PR #194, opened 2024-01-23
Fix runtime error introduced in SUP-1695

lizrabuya:
In a rush to fix a linting error, I didn't immediately realise that the following code block introduced a runtime error:

```
err := browser.OpenURL(*agent.WebURL)
return m, m.agentList.NewStatusMessage(fmt.Sprintf("Failed opening agent web url: %s", err.Error()))
```

If `browser.OpenURL()` does not fail, then `err.Error()` results to a runtime error.  This PR aims to fix that.

### DIFF
--- a/internal/agent/list.go
+++ b/internal/agent/list.go
@@ -86,8 +86,9 @@ func (m AgentListModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 		case "w":
 			if agent, ok := m.agentList.SelectedItem().(AgentListItem); ok {
-				err := browser.OpenURL(*agent.WebURL)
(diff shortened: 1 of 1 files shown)


james2791 (approved):

mcncl (changes requested):
A brief description of the change would be good, just so we don't need to go to the file changes in the future to work out what was done.

jradtilbrook (approved):

lizrabuya:
> A brief description of the change would be good, just so we don't need to go to the file changes in the future to work out what was done.

Ah yes.. the PR was done in a rush (after I panicked when discovering it) so I forgot about that 😅

mcncl (approved):

PR #195, opened 2024-01-29
SUP-1755 Add keybindings type

jradtilbrook:
This new package is intended to be used by a list component for configuring actions when a user presses a key

It builds upon the idea of keymaps in the bubbles list model by adding the callback to execute when the key is pressed. It also implements an interface that makes it renderable by the bubbles help model as well so it will "just work" when we put it into another model


lizrabuya (approved):

james2791 (approved):
:shipit:

PR #199, opened 2024-02-06
Revert "SUP-1693 Bind "s" to stop agent"

james2791:
Reverts buildkite/cli#188

Looking like the agent stop keybind (`s`|`S`) broke `bk agent list` on proceeding PRs; rolling this back for now to get attention to the cause (`tea.windowSizeMsg` alter?)

Edit - Linear also auto re-opened with integration


mcncl (approved):
It would be good to revisit the implementation on keybinding agent stopping.

PR #202, opened 2024-02-08
SUP-1751 bk build rebuild implementation

lizrabuya:
This PR implements the command `bk build rebuild <number> <pipeline> -w`, calling the [rebuild a build api ](https://buildkite.com/docs/apis/rest-api/builds#rebuild-a-build) from the specified build number of a pipeline and returns the url of the newly created build. 

The -w flag is also implemented together with this PR where a browser is opened for the created build.

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -33,6 +33,7 @@ func NewCmdBuild(f *factory.Factory) *cobra.Command {
 
 	cmd.AddCommand(NewCmdBuildNew(f))
 	cmd.AddCommand(NewCmdBuildView(f))
+	cmd.AddCommand(NewCmdBuildRebuild(f))
 
 	return &cmd
 }
--- a/pkg/cmd/build/rebuild.go
+++ b/pkg/cmd/build/rebuild.go
@@ -1,0 +1,66 @@
+package build
+
+import (
+	"fmt"
+
+	"github.com/MakeNowJust/heredoc"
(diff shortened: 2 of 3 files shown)


mcncl (approved):
This looks good. Perhaps something we can consider in the future is accepting a single arg which is the URL of the build being retried; at the moment it needs to be `bk build rebuild 5 org/pipeline` which feels long?

Maybe we can look in to `bk build rebuild https://buildkite.com/org/pipeline/builds/5`

I'd maybe also consider creating an alias so `bk rebuild` maps to `bk build rebuild`.

lizrabuya:
Thanks for the feedback @mcncl. I will create linear tickets for those 👍

lizrabuya:
> This looks good. Perhaps something we can consider in the future is accepting a single arg which is the URL of the build being retried; at the moment it needs to be `bk build rebuild 5 org/pipeline` which feels long?
> 
> Maybe we can look in to `bk build rebuild https://buildkite.com/org/pipeline/builds/5`
> 
> I'd maybe also consider creating an alias so `bk rebuild` maps to `bk build rebuild`.

@mcncl, I just realised that `bk build rebuild https://buildkite.com/org/pipeline/builds/5` might no longer be needed once we have implement bk build to resolve to a default pipeline. When this happens the command will now be `bk build rebuild 5` or `bk rebuild 5` (https://linear.app/buildkite/issue/SUP-1814/bk-rebuild-alias-to-bk-build-rebuild)

PR #214, opened 2024-03-11
SUP-1911 Convert build command pipeline resolution

jradtilbrook:
This converts current build commands to an `AggregateResolver` implementation. Currently it only resolves using the existing positional argument but we can append to the resolver as we implement them

### DIFF
--- a/internal/pipeline/resolver.go
+++ b/internal/pipeline/resolver.go
@@ -1,19 +1,24 @@
 package pipeline
 
-// TODO temporary type for resolvers to return. this should be amended in future work
-type Pipeline any
+import "errors"
+
+// Pipeline is a struct containing information about a pipeline for a resolver to return
+type Pipeline struct {
+	Name string
+	Org  string
+}
 
 // PipelineResolverFn is a function for the purpose of finding a pipeline. It returns an error if an irrecoverable
 // scenario happens and should halt execution. Otherwise if the resolver does not find a pipeline, it should return
 // (nil, nil) to indicate this. ie. no error occurred, but no pipeline was found either.
-type PipelineResolverFn func() (Pipeline, error)
+type PipelineResolverFn func() (*Pipeline, error)
 
 type AggregateResolver []PipelineResolverFn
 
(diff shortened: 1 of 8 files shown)


internal/pipeline/resolver.go
```diff
@@ -1,19 +1,24 @@
...
 
-// TODO temporary type for resolvers to return. this should be amended in future work
-type Pipeline any
+import "errors"
+
+// Pipeline is a struct containing information about a pipeline for a resolver to return
+type Pipeline struct {
+	Name string
+	Org  string
+}
```
jradtilbrook, line 9: I'm not sure what to use here. A simple struct like this we can add fields to as necessary? Or should we use an existing struct from the go-buildkite package or something? (I think a specific dependency on that would not be ideal)

pkg/cmd/build/new.go
```diff
@@ -20,17 +22,21 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 
 	cmd := cobra.Command{
 		DisableFlagsInUseLine: true,
-		Use:                   "new <pipeline> [flags]",
+		Use:                   "new [pipeline] [flags]",
 		Short:                 "Creates a new pipeline build",
-		Args:                  cobra.ExactArgs(1),
+		Args:                  cobra.MaximumNArgs(1),
```
jradtilbrook, line 27: I've changed this here/now because more resolvers are coming shortly which will make this argument optional. For now it is actually required though but it seemed like an okay thing to do

pkg/cmd/build/new.go
```diff
@@ -42,6 +48,24 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 	return &cmd
 }
 
+func pipelineResolverPositionArg(args []string, conf *config.Config) pipeline.PipelineResolverFn {
```
jradtilbrook, line 51: I think these functions could be better or live somewhere else. This serves as the idea for how to integrate it at least

mcncl (approved):

PR #216, opened 2024-03-26
Use .bk.yaml universally for config

mcncl:
Currently we use either `.bk.yaml` in `~/.config` or `buildkite.yaml` in the repository, but the latter would clash with a pipeline file.

We should use a consistent name for the configuration file.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,7 +16,7 @@ const (
 
 const (
 	appData        = "AppData"
-	configFilePath = ".bk.yaml"
+	configFilePath = "bk.yaml"
 	xdgConfigHome  = "XDG_CONFIG_HOME"
 )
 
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -16,7 +16,7 @@ const (
 
 const (
 	appData        = "AppData"
-	configFilePath = "bk.yaml"
+	configFilePath = ".bk.yaml"
```
jradtilbrook, line 19: When using the global config file, we don't want the file to be hidden. So this will need some further changes I think.

Or possible just revert this change, since it looks hard coded below for the "local" config

jradtilbrook (changes requested):

jradtilbrook (approved):

PR #217, opened 2024-03-26
Initial framework for adding an OpenAI token to the base config

mcncl:
We don't need to have the functionality **yet**, but the initial field could be set on first config right now.

We'll need to check for its existence to replace the empty string in the future.

I can see a future where each org may have its own token, but at the moment a top level makes sense to me; we can always change it later.


jradtilbrook (approved):
👌

PR #222, opened 2024-04-17
SUP-2009 | Fix `build` commands

mcncl:
## Change
All `build` commands have been parsing the `args` incorrectly since the introduction of the resolvers.

The first (0) `arg` in the list is always the build number, `buildId` in code, so we need not pass that to the resolver.

We know that then the next command has to be the pipeline in either `[ORG/PIPELINE]` or just `[PIPELINE]` format.

### Additional
Removed any reference to using the full URL to find the build; that doesn't parse correctly at the moment and we could do some `regexp` magic to make it nicer anyway.


lizrabuya (approved):
👍

PR #225, opened 2024-04-25
SUP-1903 Add path based pipeline resolver

jradtilbrook:
This adds a resolver that uses the current path and API results to determine a pipeline automatically for the user.

### DIFF
--- a/internal/artifact/artifact.go
+++ b/internal/artifact/artifact.go
@@ -1,0 +1,24 @@
+package artifact
+
+import "fmt"
+
+func FormatBytes(bytes int64) string {
+	const (
+		KB = 1024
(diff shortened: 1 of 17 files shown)


mcncl:
I'm not sold on the resolvers being in the `pipeline` package, especially in things like `build` commands where you aren't really resolving a pipeline. 

I think there should be a `resolver(s)` package within `internal` that houses the various different types of `resolver`. For examples when we eventually have a _cluster_ `resolver` will that be contained within the `pipeline` package?

jradtilbrook:
I agree @mcncl 👍 

I've made a change to the structure now, what do you think?

mcncl (approved):

PR #227, opened 2024-04-29
Add annotation outputting

mcncl:
## Changes
This allows annotations to be outputted with the `build view` command, the caveat with annotation output is that they are used in weird and wonderful ways;
* Test outputs
* Snyk results
* etc

Therefore it makes sense to set a limit on the body length of the annotations, displaying a hint (`...(more)`) that there is additional information not being shown.

CleanShot 2024-04-29 at 10 29 48

### DIFF
--- a/internal/annotation/annotation.go
+++ b/internal/annotation/annotation.go
@@ -1,0 +1,33 @@
+package annotation
+
+import (
+	"regexp"
+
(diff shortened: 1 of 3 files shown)


jradtilbrook (approved):
Looks good

PR #229, opened 2024-04-30
SUP-1596 Configure goreleaser to create github releases

jradtilbrook:
This updates the goreleaser config to allow it to create github releases.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -8,6 +8,9 @@ release:
   make_latest: false
   mode: replace
 
+changelog:
+  use: github
+
 builds:
   - id: macos
     goos: [darwin]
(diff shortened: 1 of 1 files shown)


PriyaSudip (approved):

jradtilbrook:
I used this config to create a release against my fork and it looks like its what we want: https://github.com/jradtilbrook/bk-cli/releases

mcncl (approved):
Nice!

Do we want it to remain in draft?

lizrabuya (approved):

jradtilbrook:
> Nice!
> 
> Do we want it to remain in draft?

@mcncl I'm thinking yes just so we can have a final check over it. At least for now, and when we get more confident and automated, maybe then we can just let it go for it straight awy

PR #234, opened 2024-04-30
SUP-1761 Implement local config pipeline resolver func

lizrabuya:
- **SUP-1453 Allow local directory config override for default pipeline**
- **Add pipeline resolver for local config**
- **Update git package version**
- **update packages**
- **local config can have comma separated default pipelines**
- **Add pipeline resolver config**
- **fix lint and test errors**
- **fix merge issues**
- **Add local config to gitignore**


pkg/cmd/build/view.go
```diff
@@ -41,6 +41,7 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 			resolvers := resolver.NewAggregateResolver(
 				resolver.ResolveFromPositionalArgument(args, 1, f.Config),
 				resolver.ResolveFromPath("", f.Config.Organization, f.RestAPIClient),
+				resolver.ResolveFromConfig(f.LocalConfig),
```
jradtilbrook, line 44: I think we want these above the ResolveFromPath

internal/pipeline/resolver/config.go
```diff
@@ -1,25 +1,45 @@
...
+		defaultExists := false
+		for _, opt := range c.Pipelines {
+			if defaultPipeline == opt {
+				defaultExists = true
+			}
+			pipelines = append(pipelines, opt)
+		}
+
+		if !defaultExists { //add default pipeline to the list of pipelines
+			pipelines = append([]string{defaultPipeline}, pipelines...)
```
jradtilbrook, line 32:
```suggestion
			pipelines = append(pipelines, defaultPipeline)
```

jradtilbrook (commented):
I think looks good, but the order of the resolvers seems wrong. I think we want to use the config file before we try use the API

internal/pipeline/resolver/config.go
lizrabuya: @jradtilbrook , this was intentional so that the default pipeline will always be first on the list when prompting to the user. But my renderer implementation already took care by always selecting the default 😅

jradtilbrook: ohh right okay 👍

pkg/cmd/build/view.go
jradtilbrook: Need to apply this to all the commands

lizrabuya: missed to save the other commands and didn't double check the commit 😓

jradtilbrook (approved):

PR #235, opened 2024-04-30
SUP-2141 Add build resolver structure

jradtilbrook:
Test out some structure for build resolution

### DIFF
--- a/internal/build/resolver/resolver.go
+++ b/internal/build/resolver/resolver.go
@@ -1,0 +1,5 @@
+package resolver
+
+import "github.com/buildkite/cli/v3/internal/build"
+
+type BuildResolverFn func() (*build.Build, error)
--- a/internal/build/resolver/resolver.go
+++ b/internal/build/resolver/resolver.go
@@ -7,10 +7,6 @@ import (
 	"github.com/buildkite/cli/v3/internal/build"
 )
 
-var resolved bool
-var resolvedBuild *build.Build
-var resolvedError error
-
 // BuildResolverFn is a function for finding a build. It returns an error if an irrecoverable scenario happens and
 // should halt execution. Otherwise, if the resolver does not find a build, it should return (nil, nil) to indicate
 // this. ie. no error occurred, but no build was found either
@@ -24,21 +20,12 @@ type AggregateResolver []BuildResolverFn
 //
 // This is safe to call multiple times, the same result will be returned
 func (ar AggregateResolver) Resolve(ctx context.Context) (*build.Build, error) {
-	// short-circuit if this has been called before
-	if resolved {
-		return resolvedBuild, resolvedError
-	}
-
-	resolved = true
-
 	for _, resolve := range ar {
 		b, err := resolve(ctx)
 		if err != nil {
-			resolvedError = err
 			return nil, err
 		}
 		if b != nil {
-			resolvedBuild = b
 			return b, nil
 		}
 	}
--- a/internal/build/resolver/resolver.go
+++ b/internal/build/resolver/resolver.go
@@ -1,5 +1,60 @@
 package resolver
 
-import "github.com/buildkite/cli/v3/internal/build"
+import (
+	"context"
+	"errors"
 
-type BuildResolverFn func() (*build.Build, error)
+	"github.com/buildkite/cli/v3/internal/build"
+	"github.com/buildkite/cli/v3/internal/pipeline"
+)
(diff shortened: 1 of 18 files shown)


internal/build/resolver/resolver.go
```diff
@@ -0,0 +1,60 @@
...
+var resolved bool
+var resolvedBuild *build.Build
+var resolvedError error
+
+// BuildResolverFn is a function for finding a build. It returns an error if an irrecoverable scenario happens and
+// should halt execution. Otherwise, if the resolver does not find a build, it should return (nil, nil) to indicate
+// this. ie. no error occurred, but no build was found either
+//
+// Some build resolvers may need to limit scope to a pipeline so that is provided to the resolver at runtime
+type BuildResolverFn func(context.Context, pipeline.Pipeline) (*build.Build, error)
```
jradtilbrook, line 20: Do we like the ergonomics of this? Its similar to the pipeline resolver but we should start passing around context (for use in things like API requests). And some build resolvers will need to know the resolved pipeline, should that be passed in here or passed by closure like the other pipeline resolver functions?

internal/build/resolver/resolver.go
```diff
@@ -0,0 +1,60 @@
...
+type AggregateResolver []BuildResolverFn
+
+// Resolve is a BuildResolverFn that wraps up a list of resolvers to loop through and try find a build. The first build
+// to be found will be returned. If none are found, it won't return an error to match the expectation of a
+// BuildResolverFn
+//
+// This is safe to call multiple times, the same result will be returned
+func (ar AggregateResolver) Resolve(ctx context.Context, p pipeline.Pipeline) (*build.Build, error) {
+	// short-circuit if this has been called before
+	if resolved {
```
jradtilbrook, line 31: We could do something like this for the pipeline resolver as well so that its safe to call multiple times and it gives the same result.

That way, instead of a build resolver accepting a fully resolved pipeline, it could accept a PipelineResolverFn and call that to get its pipeline (if it needs one) and if its been called before it won't do all the API/IO logic again

internal/build/resolver/resolver.go
jradtilbrook: Or building on the comment below, this could accept a PipelineResolverFn that is cached instead: 
```suggestion
type BuildResolverFn func(context.Context, pipelineresolver.PipelineResolverFn) (*build.Build, error)
```

jradtilbrook:
All this is to achieve a JIT resolution of everything the CLI will need, we can continue extrapolating this out for other resources if/when necessary

mcncl:
What is the context going to be used for here? I'm assuming it's not to pass data around but rather to use things like `cancel`?

jradtilbrook:
> What is the context going to be used for here? I'm assuming it's not to pass data around but rather to use things like `cancel`?

No, more intended for things like cancelling API requests, etc

internal/build/resolver/resolver.go
jradtilbrook: I opted to remove this entirely from the definition of a resolver. If a resolver has another kind of dependency, such as needing to resolve a pipeline, it should be provided by binding it to a closure instead

internal/build/resolver/resolver.go
jradtilbrook: I think this might be a good idea. I've added it to the pipeline as well. I'll still add a test against it as well though

internal/build/resolver/url.go
```diff
@@ -1,14 +1,15 @@
...
 	"fmt"
 	"regexp"
 
 	"github.com/buildkite/cli/v3/internal/build"
 )
 
-func ResolveFromURL(args []string) build.BuildResolverFn {
-	return func() (*build.Build, error) {
+func ResolveFromURL(args []string) BuildResolverFn {
+	return func(context.Context) (*build.Build, error) {
```
jradtilbrook, line 12: Eg, if this resolver needed to know the pipeline to be able to resolve the build (ie thinking "most recent build on a branch for a pipeline"), we would bind in a PipelineResolverFn here for the function to use. The signature of the build resolver would remain the same and we don't need to expose that dependency to other resolvers that may not need it (like this one)

jradtilbrook: The API request that is made to determine the "most recent build" would also make use of the context passed in so that it could cancel the request
p

lizrabuya (approved):

PR #237, opened 2024-05-03
SUP-2158 Add genqlient for GraphQL API

jradtilbrook:
Add and configure genqlient for making GraphQL requests


generate.go
```diff
@@ -0,0 +1,39 @@
...
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		if err = os.WriteFile(schemaFile, []byte(schema), 0644); err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		fmt.Printf("Written schema to %s\n", schemaFile)
```
mcncl, line 34: I think `Wrote` is more correct, or `Schema written to...`

mcncl (changes requested):

generate.go
jradtilbrook:
```suggestion
		fmt.Printf("Schema written to %s\n", schemaFile)
```

mcncl (approved):

PR #243, opened 2024-05-08
SUP-2138 Refactor config package

jradtilbrook:
This started out to see how we could better manage the configuration for the CLI a little nicer in some sort of hierarchy and ended up quite big. I think where its landed is quite ideal though.

With this change, the configuration follows similar to other Unix configuration files where you have a local config (current directory), a user config (somewhere under $HOME), and potentially down the track a system config (under /etc).

Configuration precedence follows that order as well. So the configuration file "closest" to where `bk` is invoked has the highest precedence.

This will mean we can remove the LocalConfig object (created another card already).

The file structure was not changed. But access patterns have been improved. 

This has also resulted in much less code with more deletions than additions (and that doesn't factor in the followup to remove LocalConfig).


jradtilbrook:
Tests are failing locally for me. So we appear to have a flaky test. I believe it has something to do with `gock` as it mentions its not really meant for current use (in our parallel tests).

I'll spend some time to try fix it, but at the least, I'll remove the parallel testing if it takes too long

lizrabuya (approved):

PriyaSudip (commented):
Reviewing

PriyaSudip (approved):

jradtilbrook:
I've opted to remove gock and go with a stubbed RoundTripper for the tests instead. That allows them to work concurrently

lizrabuya (approved):

PR #244, opened 2024-05-09
Add a type for PipelineChooser

jradtilbrook:
This type will be used by resolvers to allow them to reduce multiple pipelines they find down to a single one


mcncl (approved):
```go
type PickedPipeline pipeline.Pipeline

type PipelinePicker func([] PickedPipeline) *PickedPipeline
```

PR #245, opened 2024-05-09
Add PipelinePicker to ResolveFromRepository

jradtilbrook:
This introduces a PipelinePicker to the ResolveFromRepository resolver to allow picking a pipeline from the multiple returned from the API. It doesn't implement a PipelinePicker at the moment and just uses a simple one that hardcodes the first entry. Other picker logic/implementations will be added in subsequent PRs

### DIFF
--- a/internal/pipeline/resolver/picker.go
+++ b/internal/pipeline/resolver/picker.go
@@ -7,6 +7,9 @@ import (
 	"github.com/buildkite/cli/v3/internal/pipeline"
 )
 
+// PipelinePicker is a function used to pick a pipeline from a list.
+//
+// It is indended to be used from pipeline resolvers that resolve multiple pipelines.
 type PipelinePicker func([]pipeline.Pipeline) *pipeline.Pipeline
 
 func PassthruPicker(p []pipeline.Pipeline) *pipeline.Pipeline {
--- a/internal/pipeline/resolver/config.go
+++ b/internal/pipeline/resolver/config.go
@@ -7,14 +7,14 @@ import (
 	"github.com/buildkite/cli/v3/internal/pipeline"
 )
 
-func ResolveFromConfig(conf *config.Config) PipelineResolverFn {
+func ResolveFromConfig(conf *config.Config, picker PipelinePicker) PipelineResolverFn {
 	return func(context.Context) (*pipeline.Pipeline, error) {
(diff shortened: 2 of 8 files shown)


internal/config/config.go
```diff
@@ -126,6 +127,17 @@ func (conf *Config) HasConfiguredOrganization(slug string) bool {
...
 
+func (conf *Config) SavePipelines(pipelines []pipeline.Pipeline) error {
+	names := make([]string, len(pipelines))
+	for i, n := range pipelines {
+		names[i] = n.Name
+	}
+
+	conf.localConfig.Set("pipelines", names)
+
+	return conf.localConfig.WriteConfig()
```
mcncl, line 138: We can `return` early from this by checking the `len > 0` at the beginning of the function, it will also avoid setting a `pipelines` key where not needed.

internal/pipeline/resolver/picker.go
```diff
@@ -4,4 +4,11 @@ import (
...
 )
 
+// PipelinePicker is a function used to pick a pipeline from a list.
+//
+// It is indended to be used from pipeline resolvers that resolve multiple pipelines.
 type PipelinePicker func([]pipeline.Pipeline) *pipeline.Pipeline
+
+func PassthruPicker(p []pipeline.Pipeline) *pipeline.Pipeline {
+	return &p[0]
+}
```
mcncl, line 14: This is the temporary fix I assume?

mcncl (changes requested):

internal/pipeline/resolver/picker.go
jradtilbrook: Yeah and maybe just for tests as well. Intentionally basic (and not checking nil) just to show example usage. We probably should delete this once we have "real" pickers

internal/config/config.go
jradtilbrook: Oh thats good point. In that case it would write an empty list to the config which would resolve first and then never go out to the API to load more either so that makes sense to not "cache" an empty list I think 👍

jradtilbrook: I think I'll close this PR. And I've applied this fix in #249

PriyaSudip (approved):

lizrabuya (approved):

mcncl (approved):

PR #249, opened 2024-05-09
Add CachedPicker to save pipelines to config

jradtilbrook:
Add a picker that will save resolved pipelines to local config and run a picker to choose one for the current invocation

I've also come across various places where we are trying to run tests in parallel but not configuring it quite correctly. Good news is I found a linter that will tell us when this happens. I've enabled that linter and also fixed up some of the errors it found for us

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"runtime"
 
+	"github.com/buildkite/cli/v3/internal/pipeline"
 	"github.com/go-git/go-git/v5"
 	"github.com/spf13/afero"
 	"github.com/spf13/viper"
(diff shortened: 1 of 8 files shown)


internal/pipeline/resolver/picker_test.go
```diff
@@ -44,7 +44,27 @@ func TestPickers(t *testing.T) {
...
+
+		pipelines := []pipeline.Pipeline{
+			{Name: "first"},
+			{Name: "second"},
+			{Name: "third"},
+		}
+		resolver.CachedPicker(conf, func(p []pipeline.Pipeline) *pipeline.Pipeline { return &p[1] })(pipelines)
+
+		b, _ := afero.ReadFile(fs, ".bk.yaml")
+		expected := "pipelines:\n    - second\n    - first\n    - third\n"
```
mcncl, line 67: Is this using spaces or tabs? Perhaps:

```go
"pipelines:\n\t- second\n\t- first\n\t- third\n"
```

jradtilbrook: TIL tab characters are not valid in yaml. So these are spaces

internal/config/config.go
```diff
@@ -130,13 +131,31 @@ func (conf *Config) HasConfiguredOrganization(slug string) bool {
 	return ok
 }
 
-func (conf *Config) AddPreferredPipeline(pipelines []string) error {
-	conf.localConfig.Set("preferences.pipelines", pipelines)
-	return conf.localConfig.WriteConfig()
+func (conf *Config) PreferredPipelines() []pipeline.Pipeline {
```
jradtilbrook, line 134: I changed these to take in the pipeline objects since I'd like those to be the things we pass around

internal/pipeline/resolver/config.go
```diff
@@ -2,28 +2,19 @@ package resolver
...
 
 	"github.com/buildkite/cli/v3/internal/config"
 	"github.com/buildkite/cli/v3/internal/pipeline"
 )
 
-func ResolveFromConfig(c *config.Config) PipelineResolverFn {
+func ResolveFromConfig(conf *config.Config) PipelineResolverFn {
 	return func(context.Context) (*pipeline.Pipeline, error) {
-
-		if c == nil {
```
jradtilbrook, line 14: We have a validator that will ensure config is not nil at a higher level than this so I opted to remove it

mcncl (approved):

PR #258, opened 2024-05-20
Use build resolvers in build commands

jradtilbrook:
Update the build commands to use the build resolver chain we have now. This also now makes the build argument optional since it enables the "current branch" build resolver


mcncl (approved):
For future; we should try and keep the PR simple/only changing a single function. For this it might have been good to have the branch resolver in a PR, then adding it to existing functions in another PR.

PR #260, opened 2024-05-24
Add NFPM to goreleaser

jradtilbrook:
Changes:

- introduced a Dockerfile that pulls from goreleaser
  - this allows dependabot to keep our goreleaser version up to date for us so we are always using the newest version
- consolidated the build/release portions of docker-compose to use the Dockerfile
  - both services were using the same image with different config, that config has moved to pipeline.yaml now instead
- the build step now produces packages and pushes them to artifacts. we might push them to the registry in a future PR, or leave it for only 3.x builds
- added a release.sh script for housing the building and pushing of packages
- updated .goreleaser.yaml to create linux packages
- use goreleaser pro and split the build so it runs faster

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -1,0 +1,4 @@
+# This uses the upstream image without any changes intentionally.
+# This allows dependabot to keep this version up-to-date for us automatically, since it does not work with
+# docker-compose directly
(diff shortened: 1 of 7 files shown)


lizrabuya (dismissed):

mcncl (dismissed):
Any reason to keep the `contents` in the code?

jradtilbrook:
> Any reason to keep the `contents` in the code?

@mcncl oh oops I wasn't actually finished just yet

PriyaSudip (dismissed):

lizrabuya (approved):

PR #266, opened 2024-06-03
Add checks if a remote URL exists, or if repo exists

lizrabuya:
This PR resolves the segfault issue reported by @patrobinson when running the cli in a directory without a remote set. 

Two scenarios are addressed here:

- we should check if in a repository within the ResolveFromRepository resolver prior to making API requests
_if not in a repo, we can pass through (returning nil, nil)_

- we need to also check the repo's "origin" remote exists in the Remotes map prior to retrieving the URLs field. if origin _does not exist, we can passthrough (nil, nil)_

Updated/refactored the tests for Repository Resolver function

### DIFF
--- a/internal/pipeline/resolver/repository.go
+++ b/internal/pipeline/resolver/repository.go
@@ -2,7 +2,6 @@ package resolver
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/pipeline/resolver/repository.go
```diff
@@ -82,12 +81,16 @@ func filterPipelines(repoURLs []string, org string, client *buildkite.Client) ([
 
 func getRepoURLs(r *git.Repository) ([]string, error) {
 	if r == nil {
-		return nil, fmt.Errorf("could not determine current repository")
+		return nil, nil // could not resolve to any repository, proceed to another resolver
 	}
```
jradtilbrook, line 85: @lizrabuya will this work the way we expect by checking here? It looks like it would still end up making some API calls that are redundant if we don't know the repository.

Line https://github.com/buildkite/cli/pull/266/files#diff-5fe66e65fdcc424120bfcc3ffe59abe20947c796250173ec0b13378dc4f28845R42 only checks if this returns an error, otherwise it proceeds.

I think we should check for a repository on line 20 or so probably

PR #267, opened 2024-06-03
Runtime error encountered when viewing build with no commit SHA

lizrabuya:


internal/build/view.go
```diff
@@ -16,7 +16,10 @@ func BuildSummary(build *buildkite.Build) string {
 		buildCreator(build),
 		build.CreatedAt.UTC().Format(time.RFC1123Z))
 	hash := *build.Commit
-	commitDetails := fmt.Sprintf("Branch: %s / Commit: %s \n", *build.Branch, hash[0:8])
+	if len(hash) > 8 {
```
jradtilbrook, line 19: Should this instead be:
```suggestion
	if len(hash) >= 7 {
```

jradtilbrook (approved):

PR #271, opened 2024-06-04
`bk build download --logs` command

jradtilbrook:
This command will allow downloading all job logs for a build. It creates a directory and puts each job's log into a separate file within.

### DIFF
--- a/pkg/cmd/build/download.go
+++ b/pkg/cmd/build/download.go
@@ -1,0 +1,51 @@
+package build
+
+import (
+	"fmt"
+
+	buildResolver "github.com/buildkite/cli/v3/internal/build/resolver"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdBuildDownload(f *factory.Factory) *cobra.Command {
+	var logs bool
+
+	cmd := cobra.Command{
+		DisableFlagsInUseLine: true,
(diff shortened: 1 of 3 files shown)


pkg/cmd/build/download.go
```diff
@@ -0,0 +1,102 @@
...
+	var logs bool
+
+	cmd := cobra.Command{
+		DisableFlagsInUseLine: true,
+		Use:                   "download [number [pipeline]] [flags]",
+		Short:                 "Download resources of a build",
+		Long:                  "Download allows you to download resources from a build.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !logs {
+				fmt.Println("Nothing to download")
```
mcncl, line 27: I think this message is a little inaccurate; not passing a flag doesn't mean there aren't logs to download, more that no option has been set. At the moment, this is the only `downloadable`, so we should maybe output a message around that? Eg, if I use

```
bk download 555 test-pipeline
```
It is safe to assume I intended to download _something_, so maybe we output the help?

mcncl (changes requested):

pkg/cmd/build/download.go
jradtilbrook: Yeah I struggled with that. It's almost like a flag is the wrong choice for this. Or perhaps we should switch it so that if nothing specified, it downloads everything (and only logs are implemented for now). Does that sound better?

mcncl: Maybe default to logs for now then provide helper text as we add stuff?

lizrabuya (approved):

PriyaSudip (approved):

mcncl (approved):

PR #272, opened 2024-06-05
Allow setting of AI config using ai command

mcncl:
## Changes
This adds in the functionality of adding OpenAI tokens to your **userConfig**; localConfig is not something we should apply these tokens too, in the interest of security.

The `ai configure` command will allow the user to create an OpenAI token with a key (name) and value, they can create one or many keys as it's possible to limit the scope of a key.

`bk ai use` will let them select a key to use, and maintain that as the default key unless changed.


internal/config/config.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	DefaultGraphQLEndpoint     = "https://graphql.buildkite.com/v1"
 	OrganizationsSlugConfigKey = "organizations"
+	OpenAITokenKey             = "openai_token"
```
jradtilbrook, line 24: We don't need to access things through these externally (and its not used internally anyway so we can remove it

pkg/cmd/ai/add/add.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdAIAdd(f *factory.Factory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add",
```
jradtilbrook, line 11: I think this should follow the `config` command instead which is used to add the API token for BK

pkg/cmd/ai/ai.go
```diff
@@ -0,0 +1,25 @@
...
+func NewCmdAI(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "ai <command>",
+		Short: "Manage AI integration.",
+		Long:  "Work with Buildkite AI.",
+		Example: heredoc.Doc(`
+			# To configure your AI token
+			$ bk ai configure
+		`),
+		PersistentPreRunE: validation.CheckValidConfiguration(f.Config),
```
jradtilbrook, line 19: This validation checks an API token and org are configured but its not necessary for this command so can remove it

pkg/cmd/ai/configure.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdAIConfigure(f *factory.Factory) *cobra.Command {
+	var force bool
+
+	cmd := &cobra.Command{
+		Use:     "configure",
+		Aliases: []string{"config"},
```
jradtilbrook, line 16: Ohh is this scoped to the parent command? Or will it clobber the existing "config" alias

jradtilbrook (changes requested):
Sweet thanks for picking this up. A few changes we should make though 👇

pkg/cmd/ai/configure.go
mcncl: The scoping for an alias is to the parent of it, `bk ai config` == `bk ai configure`, `bk config` does not get overwritten, nor does it _allow_ `ai` config.

pkg/cmd/ai/add/add.go
mcncl: It does?

jradtilbrook: Oh I see, I got it mixed up. So this adds `bk ai config add` right?

jradtilbrook (changes requested):
I think having multiple OpenAI tokens at the moment is too much overhead. We should go back to just the one.

I see users having a token for just the BK CLI with its necessary scopes/limits set on that token. Then for other uses of AI they would have other tokens. But not really for use within the CLI

mcncl:
My thoughts:

- Folks may want to have a token specifically for interacting with the CLI and not have it listed in their chats, in my view. Or may want to only allow the CLI `read only` scope rather than `write`

- Allowing multiple tokens doesn't mean they need to use multiple, if they run `bk ai configure` once then they'll only have 1 token. Only if they use `add` can they add to the list of tokens available

- Tokens also come with a balance. If the user has an API token that's associated with a Teams account for work then they may wish to use that, given they're not responsible for topping it up, using a single token may lead to a cost for the user or even queries being rejected because of a lack of balance available

jradtilbrook:
I think thats possible but not worth the architecture at the moment. We are still figuring out what AI will look like so want to weigh that up against the effort and maintenance

jradtilbrook (approved):

PR #273, opened 2024-06-06
Add job list in build output

mcncl:
## Changes
This allows the printing of build jobs to the terminal. We can kind of safely assume that a build will always have at least one job, but could implement tests too if eventually.

The output looks like:
CleanShot 2024-06-06 at 15 15 17

I didn't want the time to become the focus, so made it grey in colour.

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -27,7 +27,7 @@ func main() {
 			os.Exit(1)
 		}
 
-		if err = os.WriteFile(schemaFile, []byte(schema), 0644); err != nil {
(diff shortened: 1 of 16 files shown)


lizrabuya (approved):

jradtilbrook:
Ohh nice addition 👍 One thing though, I don't think this will handle non-command type steps as I can see it uses timestamps and falls back to the `Command` job field which would be nil. We can follow up with those improvements in another PR

PR #274, opened 2024-06-06
Implement bk pipeline create

lizrabuya:
`bk pipeline create` command prompts for a pipeline name and description, then tries to resolve a git repo URL in its directory path. If no repo URL is resolved, it will prompt for a repo URL.


pkg/cmd/pipeline/create.go
```diff
@@ -0,0 +1,127 @@
...
+		Long: heredoc.Doc(`
+			Creates a new pipeline in the current org and outputs the URL to the pipeline.
+
+			The pipeline can be a {pipeline_slug} or in the format {org_slug}/{pipeline_slug}.
+			If the pipeline argument is omitted, the user will be prompted for a name and description.
+		`),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var repoURL string
+
+			qs := []*survey.Question{
```
jradtilbrook, line 32: The survey package we are using was deprecated in April!! 😢 

We can keep using it for now and follow up with replacing it. On the readme it says to use bubbletea 😆 

One cool thing we'll be able to do with that is use the [huh package hide feature](https://github.com/charmbracelet/huh/blob/5b41f0b45136a7a93eaaf25894dd8bd0f44e510c/group.go#L138). That will gracefully allow us to accept flags or arguments to this command and build up a form to get the other fields from the user.

pkg/cmd/pipeline/create.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewCmdPipelineCreate(f *factory.Factory) *cobra.Command {
+
+	cmd := cobra.Command{
+		DisableFlagsInUseLine: true,
+		Use:                   "create",
+		Short:                 "Creates a new pipeline",
+		Args:                  cobra.MaximumNArgs(1),
```
jradtilbrook, line 22: This doesn't actually seem to take in an any args?
```suggestion
		Args:                  cobra.NoArgs,
```

pkg/cmd/pipeline/create.go
```diff
@@ -0,0 +1,127 @@
...
+		Long: heredoc.Doc(`
+			Creates a new pipeline in the current org and outputs the URL to the pipeline.
+
+			The pipeline can be a {pipeline_slug} or in the format {org_slug}/{pipeline_slug}.
+			If the pipeline argument is omitted, the user will be prompted for a name and description.
```
jradtilbrook, line 27: This doesn't apply as it takes no args
```suggestion
```

pkg/cmd/pipeline/create.go
```diff
@@ -0,0 +1,127 @@
...
+				}
+				err := survey.AskOne(prompt, &repoURL, survey.WithValidator(survey.Required))
+				if err != nil {
+					return err
+				}
+			} else {
+				err := survey.AskOne(&survey.Input{Message: "Repository URL:"}, &repoURL, survey.WithValidator(survey.Required))
+				if err != nil {
+					return err
+				}
```
jradtilbrook, line 64: Should this be wrapped into line 45 so its all done at once?

pkg/cmd/pipeline/create.go
lizrabuya: Oh yep.. I initially implemented this to accept a pipeline name but forgot to remove that after asking for inputs

pkg/cmd/pipeline/create.go
lizrabuya: I could not find an option that would prompt if the options provided is empty. But maybe there is a way with bubbletea.

jradtilbrook: Oh yeah it probably doesn't exist in this package. Something we can move to with bubbletea in the future maybe 👍

jradtilbrook (approved):

PriyaSudip (approved):

PR #275, opened 2024-06-10
Change bk build pipeline argument to a flag

jradtilbrook:
This changes the positional pipeline argument to a flag instead. This should make more sense to users to allow specifying which pipeline to operate on within the build commands

### DIFF
--- a/pkg/cmd/build/cancel.go
+++ b/pkg/cmd/build/cancel.go
@@ -14,28 +14,25 @@ import (
 
 func NewCmdBuildCancel(f *factory.Factory) *cobra.Command {
 	var web bool
+	var pipeline string
 
 	cmd := cobra.Command{
 		DisableFlagsInUseLine: true,
-		Use:                   "cancel [number [pipeline]] [flags]",
-		Short:                 "Cancels a build.",
+		Use:                   "cancel <number> [flags]",
+		Args:                  cobra.ExactArgs(1),
+		Short:                 "Cancel a build.",
 		Long: heredoc.Doc(`
-			Cancels the specified build.
-
-			It accepts a build number and a pipeline slug  as an argument.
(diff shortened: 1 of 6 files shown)


pkg/cmd/build/cancel.go
```diff
@@ -14,28 +14,25 @@ import (
 
 func NewCmdBuildCancel(f *factory.Factory) *cobra.Command {
 	var web bool
+	var pipeline string
 
 	cmd := cobra.Command{
 		DisableFlagsInUseLine: true,
-		Use:                   "cancel [number [pipeline]] [flags]",
-		Short:                 "Cancels a build.",
+		Use:                   "cancel <number> [flags]",
```
jradtilbrook, line 21: As this is a destructive operation and could cause issues when builds are cancelled whilst running, I think requiring an actual build number here is a good idea. That will mean the user won't unintentionally cancel a `main` branch build which could cause issues with deploys or other things.

pkg/cmd/build/new.go
```diff
@@ -14,25 +14,24 @@ import (
...
-
-			The pipeline can be a {pipeline_slug} or in the format {org_slug}/{pipeline_slug}.
-			If the pipeline argument is omitted, it will be resolved using the current directory.
+			Create a new build on a pipeline.
+			The web URL to the build will be printed to stdout.
 		`),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			resolvers := resolver.NewAggregateResolver(
-				resolver.ResolveFromPositionalArgument(args, 0, f.Config),
+				resolver.ResolveFromFlag(pipeline, f.Config),
```
jradtilbrook, line 34: There is no build argument to this command. So I thought making the pipeline name optional instead of an arg would make sense. Any thoughts?

lizrabuya: I am fine with that behaviour.

lizrabuya (approved):

PR #278, opened 2024-06-11
Remove extra spinner when viewing build

lizrabuya:
Removed the sleep timer before opening the web URL in the browser. The timer caused the extra spinner to show. And to maintain consistency with the `bk agent view` command, the cli will not return an output if the -w flag is specified.

### DIFF
--- a/pkg/cmd/build/view.go
+++ b/pkg/cmd/build/view.go
@@ -2,7 +2,6 @@ package build
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/buildkite/cli/v3/internal/annotation"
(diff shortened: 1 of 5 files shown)


pkg/cmd/build/view.go
```diff
@@ -57,8 +56,14 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 				return fmt.Errorf("could not resolve a build")
 			}
 
+			if web {
+				buildUrl := fmt.Sprintf("https://buildkite.com/%s/%s/builds/%s", bld.Organization, bld.Pipeline, args[0])
```
jradtilbrook, line 60: args won't necessarily have a value here. I think this should be `bld.BuildNumber`?

pkg/cmd/build/view.go
```diff
@@ -57,8 +56,14 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
...
+				fmt.Printf("Opening %s in your browser\n\n", buildUrl)
```
jradtilbrook, line 61: I think a single newline would be better? Seeing 2 in the output when I run it makes me think something is missing
```suggestion
				fmt.Printf("Opening %s in your browser\n", buildUrl)
```

jradtilbrook (changes requested):

jradtilbrook:
I think most of the "build" commands take in `-w`. Do those need updating as well?

jradtilbrook (commented):
That looks good. Just need to check what other commands are doing with `-w` and update them as well

lizrabuya:
> I think most of the "build" commands take in `-w`. Do those need updating as well?

The other build commands do not show an extra spinner and the commands needed to show the spinner while the API's are being invoked from the background.  But I will remove the extra spaces returned from the cli's output from the other build commands.

jradtilbrook (approved):

PR #285, opened 2024-06-14
Sup 2313/fix view

mcncl:
For now, this ensures we don't get seg faults on non-command job outputs.

### DIFF
--- a/internal/build/view.go
+++ b/internal/build/view.go
@@ -16,8 +16,8 @@ func BuildSummary(build *buildkite.Build) string {
 		buildCreator(build),
 		build.CreatedAt.UTC().Format(time.RFC1123Z))
 	hash := *build.Commit
-	if len(hash) >= 7 {
-		hash = hash[0:8]
+	if len(hash) > 0 {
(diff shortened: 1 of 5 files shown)


internal/build/view.go
```diff
@@ -16,8 +16,8 @@ func BuildSummary(build *buildkite.Build) string {
 		buildCreator(build),
 		build.CreatedAt.UTC().Format(time.RFC1123Z))
 	hash := *build.Commit
-	if len(hash) >= 7 {
```
jradtilbrook, line 19: Ohh could hash, or rather `build.Commit` actually be null even? So should check for that as well? I guess technically it could 🤷‍♂️ so might be best to check regardless

jradtilbrook (approved):

internal/build/view.go
mcncl: @jradtilbrook that isn't a change in this pr

mcncl: And Go will just return an empty string

PR #286, opened 2024-06-14
Add steps to build pipeline to easily trigger a release

jradtilbrook:
Allow creating a release from the build pipeline to ease/automate the process further

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -26,7 +26,7 @@ steps:
 
   - wait
 
-  - label: ":buildkite: build ({{matrix}})"
+  - label: ":terminal: build ({{matrix}})"
     matrix:
           - "darwin"
           - "linux"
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -49,3 +49,24 @@ steps:
...
+      label: ":package: Want to create a release from this commit?"
+      branches:
+        - 3.x
+        - main
+      blocked_state: passed
+
+  # this tags the commit with the input from the previous block step and pushes it to github
+  # that will trigger the buildkite-cli-release pipeline off the tag which will create a release in github
+  - label: ":rocket: Pushing a tag to release"
+    command: "echo git tag ..."
```
jradtilbrook, line 63: We'll actually want to put this in a script I think. We want to construct the tag value based on block step fields and verify that its newer, etc than any current tags

jradtilbrook:
This is all working now and created a draft release https://github.com/buildkite/cli/releases/tag/untagged-d75d787458bf21c52d7d

I realised we don't really _need_ to validate the input tag name because the release pipeline already only builds on the expected tag regex.

mcncl (approved):

lizrabuya (approved):

PR #289, opened 2024-06-18
Push linux packages to registries

jradtilbrook:
Push build linux packages to buildkite registry. 
Verified this by pushing them from the build pipeline, so it should work from the release pipeline when it gets there

Note: APK has some quirks so leaving that off for now.

### DIFF
--- a/.buildkite/upload-packages.sh
+++ b/.buildkite/upload-packages.sh
@@ -4,7 +4,7 @@
 # This script is used to upload packages to Buildkite registries
 #
 
-set -euo pipefail
+set -uo pipefail
 
 if [[ -z "${1}" ]]; then
     echo "Must pass in the package type: apk, deb, rpm"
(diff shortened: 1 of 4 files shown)


.buildkite/upload-packages.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/env bash
+
+#
+# This script is used to upload packages to Buildkite registries
+#
+
+set -uo pipefail
+
+if [[ -z "${1}" ]]; then
+    echo "Must pass in the package type: apk, deb, rpm"
```
mcncl, line 10: Should we remove `apk` from the error?

mcncl (changes requested):

.buildkite/upload-packages.sh
jradtilbrook:
```suggestion
    echo "Must pass in the package type: deb, rpm"
```

mcncl (approved):

PR #290, opened 2024-06-18
Improve output of [build view]

mcncl:
- Added spacing around sections to avoid getting lost between them
- Underlined the section headers to show separation from the body
- Adjusted the output of the annotations to be [path] rather than [name]

CleanShot 2024-06-18 at 20 54 15

### DIFF
--- a/internal/artifact/view.go
+++ b/internal/artifact/view.go
@@ -9,8 +9,8 @@ func ArtifactSummary(artifact *buildkite.Artifact) string {
 	artifactSummary := lipgloss.JoinVertical(lipgloss.Top,
 		lipgloss.NewStyle().Align(lipgloss.Left).Padding(0, 1).Render(),
 		lipgloss.JoinHorizontal(lipgloss.Left,
-			lipgloss.NewStyle().Width(30).Align(lipgloss.Left).Padding(0, 1).Render(*artifact.Filename),
-			lipgloss.NewStyle().Align(lipgloss.Left).Padding(0, 1).Render(FormatBytes(*artifact.FileSize)),
+			lipgloss.NewStyle().Width(60).Align(lipgloss.Left).Padding(0, 1).Render(*artifact.Path),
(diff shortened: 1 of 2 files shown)


jradtilbrook:
This looks quite nice. While you're here, is that top part the build message? Perhaps we should truncate that, it looks almost broken in that form, but I suspect thats just what the commit message says is it?

And if its not too difficult, don't show annotations that end up rendering nothing? Or create a card to follow up with that I reckon

mcncl:
@jradtilbrook the annotation can be a new card. Should only be a small.

internal/build/view.go
```diff
@@ -98,3 +100,13 @@ func renderBuildNumber(state string, number int) string {
 
 	return lipgloss.NewStyle().Foreground(stateColor).Render(fmt.Sprintf("Build #%d", number))
 }
+
+func trimMessage(msg string) string {
```
jradtilbrook, line 104: There is this from bubbletea which might be better to handle more cases: https://pkg.go.dev/github.com/charmbracelet/x/ansi#Truncate

jradtilbrook: Actually, this doesn't seem to output anything anymore? Screenshot 2024-06-19 at 08 29 03

mcncl: That feels a little bit like using Bubbletea where we don't need to? We could easily add `+ "..."` to the message if we want?

mcncl: CleanShot 2024-06-19 at 10 55 05

Seems fine for me?

mcncl: The output with Truncate is terrible
CleanShot 2024-06-19 at 10 59 13

jradtilbrook: Okay wow it seems to be a problem with my terminal (kitty). I run a pretty high line spacing and the output of that exact command is 1 line longer than a full screen. kitty seems to remove the first line from scrollback or something so that seems like a bug or configuration I'll have to look into. 

So I think this PR is good to go then

jradtilbrook (approved):

PR #292, opened 2024-06-19
Implement huh? for confirmation of new build

mcncl:
It might be nice for a user to have to confirm the creation of a new build, which is the action of the web.

CleanShot 2024-06-19 at 11 49 32

### DIFF
--- a/pkg/cmd/build/new.go
+++ b/pkg/cmd/build/new.go
@@ -46,13 +46,18 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 				return fmt.Errorf("could not resolve a pipeline")
 			}
 
-			_ = huh.NewConfirm().
+			form := huh.NewConfirm().
 				Title(fmt.Sprintf("Create new build on %s?", pipeline.Name)).
 				Affirmative("Yes!").
 				Negative("No!").
 				Value(&confirm).
-				WithTheme(huh.ThemeDracula()).
-				Run()
+				WithTheme(huh.ThemeDracula())
+
+      if confirm {
+        form.Skip()
+      } else {
+        form.Run()
+      }
 
 			if confirm {
 				return newBuild(pipeline.Org, pipeline.Name, f, message, commit, branch, web)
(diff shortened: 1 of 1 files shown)


jradtilbrook (changes requested):
I think thats a good addition 👍 
Although I think it would be good to add a flag to bypass it (for automation things). Something like `--confirm|yes` or other?

pkg/cmd/build/new.go
```diff
@@ -17,6 +19,7 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 	var commit string
 	var message string
 	var pipeline string
+	var confirm bool
```
jradtilbrook, line 22: Add a cmd flag around this as well. Then you can use the WithHiddenFunc of huh forms to optionally show the form if its not set already?

mcncl: Using --confirm will now skip the input.

pkg/cmd/build/new.go
```diff
@@ -43,7 +46,24 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
...
+				form.Skip()
+			} else {
+				_ = form.Run()
+			}
+
+			if confirm {
+				return newBuild(pipeline.Org, pipeline.Name, f, message, commit, branch, web)
+			} else {
+				return errors.New("Create was aborted.")
+			}
```
jradtilbrook, line 66: If I ctrl-c the form, it still proceeds somehow 😕

jradtilbrook (changes requested):

pkg/cmd/build/new.go
mcncl: Yeah, it does on the picker too. I think that's a tea thing.

pkg/cmd/build/new.go
```diff
@@ -54,6 +74,7 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVarP(&pipeline, "pipeline", "p", "", "The pipeline to build. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}.\n"+
 		"If omitted, it will be resolved using the current directory.",
 	)
+	cmd.Flags().BoolVarP(&confirm, "confirm", "", false, "Skip the confirmation step. Useful if being used in automation/CI")
```
jradtilbrook, line 77: I'm wondering about the terminology of the "confirm" flag. It could be interpreted as "ask for confirmation". vs maybe `--no-confirm` meaning dont ask and just do it.

Do you have other ideas we could use? I don't really like `--yes` as it doesn't have much meaning either. tf uses `--auto-approve` which I dont like that much either, but its clearer

jradtilbrook:
I created #296 as an alternative with a few fixes. And then messed up pushing so its caused a failed commit status on this PR. I think we should go with 296 anyway though

PR #293, opened 2024-06-19
include cluster view changes

PriyaSudip:
Changes added for cluster view command,

image

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -1,0 +1,15 @@
+package cluster
+
+type Cluster struct {
+	OrganizationSlug string
+	ClusterID        string
+	Queues           []Queue
+	Name             string
+	Description      string
+}
+
+type Queue struct {
+	Id    string
+	Name  string
+	Agent int
+}
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -9,7 +9,7 @@ type Cluster struct {
 }
 
 type Queue struct {
-	Id    string
-	Name  string
-	Agent int
+	Id           string
+	Name         string
+	ActiveAgents int
 }
--- a/internal/cluster/query.go
+++ b/internal/cluster/query.go
@@ -1,0 +1,46 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/buildkite/cli/v3/internal/graphql"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+)
+
+func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string, f *factory.Factory) (Cluster, error) {
+
+	q, err := graphql.GetClusterQueues(ctx, f.GraphQLClient, OrganizationSlug, ClusterID)
+
+	if err != nil {
+		fmt.Println("Unable to read Cluster Queues: ", err.Error())
+		return Cluster{}, err
+	}
+
+	ClusterDescription := q.Organization.Cluster.Description
+
+	cluster := Cluster{
+		OrganizationSlug: OrganizationSlug,
+		ClusterID:        ClusterID,
+		Name:             q.Organization.Cluster.Name,
+		Description:      string(*ClusterDescription),
+		Queues:           []Queue{},
+	}
+
+	for _, edge := range q.Organization.Cluster.Queues.Edges {
+		agent, err := graphql.GetClusterQueueAgent(ctx, f.GraphQLClient, OrganizationSlug, []string{edge.Node.Id})
+		if err != nil {
+			fmt.Println("Unable to read Cluster Queue Agents: ", err.Error())
+			return Cluster{}, err
+		}
+
+		queue := Queue{
+			Id:    edge.Node.Id,
+			Name:  edge.Node.Key,
+			Agent: len(agent.Organization.Agents.Edges),
+		}
+		cluster.Queues = append(cluster.Queues, queue)
+
+	}
+	return cluster, nil
+}
--- a/internal/cluster/query.go
+++ b/internal/cluster/query.go
@@ -8,13 +8,13 @@ import (
 	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 )
 
-func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string, f *factory.Factory) (Cluster, error) {
+func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string, f *factory.Factory) (*Cluster, error) {
 
 	q, err := graphql.GetClusterQueues(ctx, f.GraphQLClient, OrganizationSlug, ClusterID)
 
 	if err != nil {
 		fmt.Println("Unable to read Cluster Queues: ", err.Error())
-		return Cluster{}, err
+		return nil, err
 	}
 
 	ClusterDescription := q.Organization.Cluster.Description
@@ -30,8 +30,8 @@ func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string
 	for _, edge := range q.Organization.Cluster.Queues.Edges {
 		agent, err := graphql.GetClusterQueueAgent(ctx, f.GraphQLClient, OrganizationSlug, []string{edge.Node.Id})
 		if err != nil {
-			fmt.Println("Unable to read Cluster Queue Agents: ", err.Error())
-			return Cluster{}, err
+			return nil, fmt.Errorf("unable to read Cluster Queue Agents: %s", err.Error())
+
 		}
 
 		queue := Queue{
@@ -42,5 +42,5 @@ func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string
 		cluster.Queues = append(cluster.Queues, queue)
 
 	}
-	return cluster, nil
+	return &cluster, nil
 }
--- a/internal/cluster/query.go
+++ b/internal/cluster/query.go
@@ -35,9 +35,9 @@ func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string
 		}
 
 		queue := Queue{
-			Id:    edge.Node.Id,
-			Name:  edge.Node.Key,
-			Agent: len(agent.Organization.Agents.Edges),
+			Id:           edge.Node.Id,
+			Name:         edge.Node.Key,
+			ActiveAgents: len(agent.Organization.Agents.Edges),
 		}
 		cluster.Queues = append(cluster.Queues, queue)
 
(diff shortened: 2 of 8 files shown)


internal/cluster/query.go
```diff
@@ -0,0 +1,46 @@
...
+
+	cluster := Cluster{
+		OrganizationSlug: OrganizationSlug,
+		ClusterID:        ClusterID,
+		Name:             q.Organization.Cluster.Name,
+		Description:      string(*ClusterDescription),
+		Queues:           []Queue{},
+	}
+
+	for _, edge := range q.Organization.Cluster.Queues.Edges {
```
mcncl, line 30: I think an edge can be `null` if no queue are found, or it can return an empty list. Should we check that the length is `> 0`?

PriyaSudip: let me check

PriyaSudip: Just did a test. cluster gets created with single queue by default and even the default one is deleted code is not broken as it is a for loop which doesn't run.

image

mcncl: Can we output `no queues` or something if the count of queues is 0?

PriyaSudip: Yea can do

PriyaSudip: Done, added message when no queues are found,

image

mcncl (approved):

internal/cluster/query.go
mcncl: So good

internal/cluster/cluster.go
```diff
@@ -0,0 +1,15 @@
...
+	ClusterID        string
+	Queues           []Queue
+	Name             string
+	Description      string
+}
+
+type Queue struct {
+	Id    string
+	Name  string
+	Agent int
```
jradtilbrook, line 14: Is this the number of agents currently registered in the queue? I think we should name it more along the lines of that if it is

pkg/cmd/root/root.go
```diff
@@ -38,6 +39,7 @@ func NewCmdRoot(f *factory.Factory) (*cobra.Command, error) {
 	cmd.AddCommand(pipelineCmd.NewCmdPipeline(f))
 	cmd.AddCommand(useCmd.NewCmdUse(f))
 	cmd.AddCommand(versionCmd.NewCmdVersion(f))
+	cmd.AddCommand(clusterCmd.NewCmdCluster(f))
```
jradtilbrook, line 42: Can you keep these sorted alphabetically

pkg/cmd/cluster/cluster.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/MakeNowJust/heredoc"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdCluster(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "cluster <command>",
+		Short: "View cluster information",
```
jradtilbrook, line 13: This one isn't technically correct since this is the parent `cluster` command. Probably change it to the Long version

pkg/cmd/cluster/cluster.go
```diff
@@ -0,0 +1,31 @@
...
+			# To view cluster details
+			$ bk cluster view -c "cluster_id"
+		`),
+		PersistentPreRunE: validation.CheckValidConfiguration(f.Config),
+		Annotations: map[string]string{
+			"help:arguments": heredoc.Doc(`
+				Organization slug and cluster id is passed as an argument. It can be supplied in any of the following formats:
+				- "ORG_SLUG"
+			`),
+		},
```
jradtilbrook, line 26: I'm not sure what this is doing and can probably be removed?

pkg/cmd/cluster/view.go
```diff
@@ -0,0 +1,51 @@
...
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdClusterView(f *factory.Factory) *cobra.Command {
+	var clusterID string
+	cmd := cobra.Command{
+		DisableFlagsInUseLine: true,
+		Use:                   "view  [flags]",
+		Args:                  cobra.MinimumNArgs(0),
```
jradtilbrook, line 17: Should this accept an arg instead of a flag for the cluster to view? I think that lines up more with other `view` commands. ie `bk build view <id>` so we should keep it consistent and convert the flag to an argument

pkg/cmd/cluster/view.go
```diff
@@ -0,0 +1,51 @@
...
+			}, "Loading cluster information")
+
+			p := tea.NewProgram(l)
+			_, err := p.Run()
+
+			return err
+		},
+	}
+
+	cmd.Flags().StringVarP(&clusterID, "clusterID", "c", "", "ID of the cluster")
```
jradtilbrook, line 46: I think remove this as above, but update the command docs to not use camelcase and explain which ID it should be (graphql or uuid)

internal/cluster/query.go
```diff
@@ -0,0 +1,46 @@
...
+		agent, err := graphql.GetClusterQueueAgent(ctx, f.GraphQLClient, OrganizationSlug, []string{edge.Node.Id})
```
jradtilbrook, line 31: This is good. Next up for your go learning is go-routines. You can fire off all these requests in parallel to speed up the resolution. We can kick that can down the road though

jradtilbrook (changes requested):
This is looking quite good, nice work 👍 

I have a few feedback items for you 👇 
Let me know if you want to pair on them

PriyaSudip:
> Changes added for cluster view command,
> 
> image



> This is looking quite good, nice work 👍
> 
> I have a few feedback items for you 👇 Let me know if you want to pair on them

Thanks for all the review comments @jradtilbrook I will action them all now

internal/cluster/query.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+)
+
+func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string, f *factory.Factory) (Cluster, error) {
+
+	q, err := graphql.GetClusterQueues(ctx, f.GraphQLClient, OrganizationSlug, ClusterID)
+
+	if err != nil {
+		fmt.Println("Unable to read Cluster Queues: ", err.Error())
+		return Cluster{}, err
```
lizrabuya, line 17: Can we return the error with the error message using go's `fmt.Errorf` function, and also use the practice to return `nil` instead of an empty Cluster object?

internal/cluster/query.go
```diff
@@ -0,0 +1,46 @@
...
+		Name:             q.Organization.Cluster.Name,
+		Description:      string(*ClusterDescription),
+		Queues:           []Queue{},
+	}
+
+	for _, edge := range q.Organization.Cluster.Queues.Edges {
+		agent, err := graphql.GetClusterQueueAgent(ctx, f.GraphQLClient, OrganizationSlug, []string{edge.Node.Id})
+		if err != nil {
+			fmt.Println("Unable to read Cluster Queue Agents: ", err.Error())
+			return Cluster{}, err
```
lizrabuya, line 34: same comment as above in returning the error with added contexts and return `nil` too

lizrabuya (changes requested):
Just some minor changes, but overall 💪🏻💪🏻!

internal/cluster/cluster.go
PriyaSudip: Yes it is registered and running agents. Should I rename it to Active Agents?

pkg/cmd/root/root.go
PriyaSudip: Done

pkg/cmd/cluster/cluster.go
PriyaSudip: Done

pkg/cmd/cluster/cluster.go
PriyaSudip: Done

internal/cluster/query.go
PriyaSudip: Sure will have a look into go-routines.

internal/cluster/cluster.go
jradtilbrook: Maybe something like connected agent count?

internal/cluster/query.go
jradtilbrook: Not for this PR though. Just for your learning. And can start to implement it in places as you go

internal/cluster/query.go
PriyaSudip: Done

internal/cluster/query.go
PriyaSudip: Done

PriyaSudip:
> Just some minor changes requested, but overall 💪🏻💪🏻!



> Just some minor changes requested, but overall 💪🏻💪🏻!

Thanks @lizette

pkg/cmd/cluster/view.go
PriyaSudip: Updated now

pkg/cmd/cluster/view.go
PriyaSudip: Updated now

internal/cluster/view.go
```diff
@@ -0,0 +1,46 @@
...
+	"strconv"
+
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/charmbracelet/lipgloss"
+	"github.com/charmbracelet/lipgloss/table"
+)
+
+func ClusterSummary(ctx context.Context, OrganizationSlug string, ClusterID string, f *factory.Factory) (string, error) {
+
+	clusterSummary, err := QueryCluster(ctx, OrganizationSlug, ClusterID, f)
```
jradtilbrook, line 16: It's not critical now but I think this would be better as separate functions. One that retrieves cluster information from the API and one that takes that and renders a string - instead of in the same function. That would lead to easier testing as well

pkg/cmd/cluster/view.go
```diff
@@ -0,0 +1,51 @@
...
+
+			p := tea.NewProgram(l)
+			_, err := p.Run()
+
+			return err
+		},
+	}
+
+	//cmd.Flags().StringVarP(&clusterID, "clusterID", "c", "", "ID of the cluster")
+	//cmd.Flags().SortFlags = false
```
jradtilbrook, line 47: Can remove these

jradtilbrook (commented):
Just remove some comments and change the Queue.Agent field name and good to go I think

pkg/cmd/cluster/view.go
PriyaSudip: Fixed it

internal/cluster/view.go
PriyaSudip: Sure Jarryd! I can take this up as next iteration

jradtilbrook (approved):

lizrabuya (approved):
🎉

PR #294, opened 2024-06-19
Don't print empty annotations

mcncl:
CleanShot 2024-06-19 at 14 58 35

CleanShot 2024-06-19 at 14 59 01

### DIFF
--- a/pkg/cmd/build/view.go
+++ b/pkg/cmd/build/view.go
@@ -96,9 +96,16 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 					}
 				}
 				if len(buildAnnotations) > 0 {
-					summary += lipgloss.NewStyle().Bold(true).Padding(0, 1).Underline(true).Render("\n\nAnnotations")
 					for _, a := range buildAnnotations {
-						summary += annotation.AnnotationSummary(&a)
+						annotationCount := 0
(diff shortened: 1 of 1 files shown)


pkg/cmd/build/view.go
```diff
@@ -96,9 +96,16 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 					}
 				}
 				if len(buildAnnotations) > 0 {
-					summary += lipgloss.NewStyle().Bold(true).Padding(0, 1).Underline(true).Render("\n\nAnnotations")
 					for _, a := range buildAnnotations {
-						summary += annotation.AnnotationSummary(&a)
+						annotationCount := 0
+						if len(annotation.AnnotationSummary(&a)) < 230 {
```
jradtilbrook, line 101: Is this special `230` something to do with the annotation box, etc?

mcncl: Yeah, the total `len` with an empty annotation is `227`, this allows for special chars that maybe use an additional byte or two

PriyaSudip (approved):

pkg/cmd/build/view.go
jradtilbrook: Hmm, what if we used glamour to render this instead. It handles markdown, which is a superset of HTML. A quick test with a string with markdown and html anchor tags and it works nicely

PR #295, opened 2024-06-20
Build resolver for user's UUID

lizrabuya:
Implement a build resolver for a given user uuid 

- [x]  Tests added


internal/build/resolver/current_user.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+
+	"github.com/buildkite/cli/v3/internal/build"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/go-buildkite/v3/buildkite"
+)
+
+// ResolveBuildFromCurrentBranch Finds the most recent build for the branch in the current working directory
+func ResolveBuildForCurrentUser(branch string, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
```
jradtilbrook, line 13: The comment looks outdated from copy-paste

internal/build/resolver/user_builds.go
```diff
@@ -12,57 +12,43 @@ import (
...
+		p, e := pipelineResolver(ctx)
+		if p != nil {
+			pipeline = p
 		}
+		return e
+	})
+	err := g.Wait()
+	if err != nil {
+		return nil, err
+	}
```
jradtilbrook, line 33: We don't need the errgroup if there is only one go-routine so can simplify this

internal/build/resolver/user_builds.go
```diff
@@ -12,57 +12,43 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// ResolveBuildFromCurrentBranch Finds the most recent build for the branch in the current working directory
-func ResolveBuildForCurrentUser(branch string, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
-	return func(ctx context.Context) (*build.Build, error) {
-		var pipeline *pipeline.Pipeline
-		var user *buildkite.User
+// ResolveBuildForUser Finds the most recent build for the user based on the provided options
+func ResolveBuildForUser(ctx context.Context, userInfo string, opt *buildkite.BuildsListOptions, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) (*build.Build, error) {
```
jradtilbrook, line 16: I think it might be better if we didn't use the `opt *buildkite.BuildsListOptions` in the argument list to avoid exposing the underlying API options. It looks like we could change this argument to the branch string and then create the BuildsListOptions within this function

internal/build/resolver/current_userid.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"github.com/buildkite/cli/v3/internal/build"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/go-buildkite/v3/buildkite"
+)
+
+func ResolveBuildFromCurrentUserId(uuid string, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
```
jradtilbrook, line 12: This isn't necessarily the current users ID. It could be any user so we should remove that from the name

jradtilbrook (changes requested):
Looks pretty good 👍 
Just a few architecture changes

internal/build/resolver/current_user.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+
+	"github.com/buildkite/cli/v3/internal/build"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/go-buildkite/v3/buildkite"
+)
+
+func ResolveBuildForCurrentUser(branch string, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
```
jradtilbrook, line 12: I think still having docs for each function is a good idea

internal/build/resolver/user_builds.go
```diff
@@ -5,64 +5,47 @@ import (
...
-			BuildNumber:  *builds[0].Number,
-		}, nil
+	if err != nil {
+		return nil, err
 	}
+	if len(builds) == 0 {
+		// we error here because this resolver is explicitly used so any case where it doesn't resolve a build is a
+		// problem
+		return nil, fmt.Errorf("failed to find a build for current user (%s)", userInfo)
+	}
```
jradtilbrook, line 44: I wonder if this applies now with these other resolvers? I suppose if the user doesn't have any builds its not really an error, we could just show a helpful message without it being an error

internal/build/resolver/userid.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+
+	"github.com/buildkite/cli/v3/internal/build"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+)
+
+func ResolveBuildByUserId(uuid string, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
```
jradtilbrook, line 11: I think it would be good to keep the naming pattern `ResolveBuildForUserID`

internal/build/resolver/user_builds.go
lizrabuya: I guess thats right, and for consistency too as we do the same for pipelines resolver. I'll update this to not error when no builds are found

jradtilbrook: actually its probably more nuanced than that. the `--user` resolver might want to error if that user doesnt exist vs. if the user does exist but doesnt have any builds

lizrabuya: @jradtilbrook , i will raise a different PR to implement this as it will have to be the same behavior for all build resolvers and ensure that this is handled as expected.

lizrabuya: if the `user` does not exist, or is invalid the api returns an error, which is also returned by the resolver. So that should be covered.

jradtilbrook (approved):
Looks good. @lizrabuya can you also create a card to come back and add this resolver to the build commands (its not used just yet but thats intentional)

PR #297, opened 2024-06-21
SUP-2363 Fix segfault when not in a repo

jradtilbrook:
There is a segfault when running `bk build view` outside of a repo. This fixes that by checking for a nil repo before usage

### DIFF
--- a/internal/build/resolver/current_branch.go
+++ b/internal/build/resolver/current_branch.go
@@ -13,6 +13,12 @@ import (
 
 // ResolveBuildFromCurrentBranch Finds the most recent build for the branch in the current working directory
 func ResolveBuildFromCurrentBranch(repo *git.Repository, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
+	// there is nothing to resolve from if we aren't in a git repository, so short circuit that
+	if repo == nil {
+		return func(ctx context.Context) (*build.Build, error) {
+			return nil, nil
+		}
+	}
 	return func(ctx context.Context) (*build.Build, error) {
 		// find the pipeline first, then we can make a graphql query to find the most recent build for that branch
 		pipeline, err := pipelineResolver(ctx)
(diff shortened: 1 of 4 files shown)


internal/build/resolver/current_branch.go
```diff
@@ -13,6 +13,12 @@ import (
 
 // ResolveBuildFromCurrentBranch Finds the most recent build for the branch in the current working directory
 func ResolveBuildFromCurrentBranch(repo *git.Repository, pipelineResolver pipelineResolver.PipelineResolverFn, f *factory.Factory) BuildResolverFn {
+	// there is nothing to resolve from if we aren't in a git repository, so short circuit that
+	if repo == nil {
+		return func(ctx context.Context) (*build.Build, error) {
+			return nil, nil
+		}
+	}
```
jradtilbrook, line 21: This is the first option. We would include this resolver every time, but it will check for a nil repo which means not in a repository, and return a dummy resolver to pass through to the next one.

internal/build/resolver/resolver.go
```diff
@@ -33,6 +33,13 @@ func (ar AggregateResolver) Resolve(ctx context.Context) (*build.Build, error) {
 	return nil, nil
 }
 
+func (ar AggregateResolver) WithResolverWhen(condition bool, resovler BuildResolverFn) AggregateResolver {
+	if condition {
+		return append(ar, resovler)
+	}
+	return ar
+}
+
```
jradtilbrook, line 42: This is the second option, see below as well. It allows chaining to include a resolver depending on the evaluation of a condition. So we could move the inclusion of the resolver out to the command level and only use it if the repo is not nil

PriyaSudip (approved):

jradtilbrook (commented):
This includes 2 options for tackling this. I'm not sure which one I like more so take a look and leave your thoughts

lizrabuya (approved):

internal/build/resolver/resolver.go
lizrabuya: I prefer this option!

internal/build/resolver/current_branch.go
lizrabuya: This approach handles the error at the right time which I think should be the right thing to do.

internal/build/resolver/resolver.go
jradtilbrook: Oh good point. So we need the first option regardless really - just to make absolute sure

PR #298, opened 2024-06-21
SUP-2321 Accept field data for `bk job unblock`

jradtilbrook:
This allows reading field data from stdin or a `--data` flag. It doesn't do any validation and leaves that up to the API, although the API only accepts a JSON formatted string.


lizrabuya (approved):
👍

PR #304, opened 2024-06-25
include go routines

PriyaSudip:
Adding go routines to run queue agent querying in parallel. 

With agents:
image
Without queue:
image


internal/cluster/query.go
```diff
@@ -27,18 +28,37 @@ func QueryCluster(ctx context.Context, OrganizationSlug string, ClusterID string
 		Queues:           []Queue{},
 	}
 
+	queueChannel := make(chan Queue, len(q.Organization.Cluster.Queues.Edges))
+	errorChannel := make(chan error, len(q.Organization.Cluster.Queues.Edges))
```
mcncl, line 32: This error channel doesn't seem to ever be read from?

mcncl (changes requested):
I wonder if we should be using more `context.WithTimeout` in our code to handle long API call issues.

internal/cluster/query.go
PriyaSudip: What would be the best place to send this information?

jradtilbrook (changes requested):
This looks okay but its all done manually. This page https://gobyexample.com/waitgroups links out to the errgroup package which will make this much simpler: https://pkg.go.dev/golang.org/x/sync/errgroup

If you search the codebase, you'll see an example of it already in use as well I think

internal/cluster/query.go
PriyaSudip: Used errgroup to handle this

mcncl (approved):

PriyaSudip:
> This looks okay but its all done manually. This page https://gobyexample.com/waitgroups links out to the errgroup package which will make this much simpler: https://pkg.go.dev/golang.org/x/sync/errgroup
> 
> If you search the codebase, you'll see an example of it already in use as well I think

Thanks @jradtilbrook as discussed I have made changes to include err group and instead of channel added array of queues and go routines based on the queue index

jradtilbrook (approved):
Sweet 👍

PR #305, opened 2024-06-26
Add homebrew info to goreleaser

jradtilbrook:
This configures goreleaser to push a formula file to https://github.com/buildkite/homebrew-buildkite

It will create a `bk@3.rb` file meaning that users can `brew install buildkite/buildkite/bk@3`.

Note: the token is not configured with access to that repository just yet, but we'll add that soon.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -12,6 +12,26 @@ release:
 changelog:
   use: github
 
+brews:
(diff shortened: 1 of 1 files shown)


PriyaSudip (approved):

jradtilbrook:
Token now has access so we can merge this

PR #310, opened 2024-07-04
Include short description for cluster

PriyaSudip:


pkg/cmd/cluster/cluster.go
```diff
@@ -9,9 +9,10 @@ import (
 
 func NewCmdCluster(f *factory.Factory) *cobra.Command {
 	cmd := cobra.Command{
-		Use:  "cluster <command>",
-		Args: cobra.ArbitraryArgs,
-		Long: "View cluster information",
+		Use:   "cluster <command>",
+		Args:  cobra.ArbitraryArgs,
+		Long:  "View cluster information",
+		Short: "View cluster and queue information",
```
mcncl, line 15: Perhaps these could be switched? Just by going off of the length of them + their identifier. Non blocker.

mcncl (dismissed):

pkg/cmd/cluster/cluster.go
jradtilbrook, line 15: Haha its funny the short version is longer than the long. Should they be switched around? Or maybe just the same thing? (queue might end up being a separate command at this stage so perhaps shouldnt include that in the cluster description)

jradtilbrook (changes requested):

pkg/cmd/cluster/cluster.go
PriyaSudip: Agree, changed it to be same as queue might end up as separate command per Jarryd.

PriyaSudip: Yes it is :D , fixed it now to be the same

pkg/cmd/cluster/cluster.go
```diff
@@ -9,9 +9,10 @@ import (
 
 func NewCmdCluster(f *factory.Factory) *cobra.Command {
 	cmd := cobra.Command{
-		Use:  "cluster <command>",
-		Args: cobra.ArbitraryArgs,
-		Long: "View cluster information",
+		Use:   "cluster <command>",
+		Args:  cobra.ArbitraryArgs,
+		Long:  "View cluster information",
+		Short: "View cluster information",
```
jradtilbrook, line 15: Oh hold on, the others are "manage" because it has things like build create/cancel/rebuild.
This will likely have more than just "view" as well so maybe?:
```suggestion
		Short: "Manage organization clusters",
```

jradtilbrook (approved):

lizrabuya (dismissed):

jradtilbrook (approved):

PR #311, opened 2024-07-08
WIP using gptscript for AI integration

jradtilbrook:


jradtilbrook:
created this just to keep track of it while producing another implementation. can close it

PR #312, opened 2024-07-09
Add documentation tool to AI command

jradtilbrook:
This update the `bk ai ask` command to make use of [tools](https://platform.openai.com/docs/guides/function-calling) ([api ref](https://platform.openai.com/docs/api-reference/chat/object)). This allows chatgpt to tell us if we should give it some more context or run something out of band.

This PR adds a tool that searches our docs algolia index for relevant pages and returns the URLs. That allows gpt-4o to parse those URLs and reference them in the response, without using extra tokens


jradtilbrook:
I think this might actually be fully ready for review

PriyaSudip (approved):

lizrabuya (approved):

jradtilbrook:
Closing this for now to come back to it when time permits

PR #316, opened 2024-07-11
Remove pending command from build cancel/rebuild/new

jradtilbrook:
Remove pending command and use simple huh? spinner

### DIFF
--- a/pkg/cmd/build/new.go
+++ b/pkg/cmd/build/new.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/MakeNowJust/heredoc"
 	"github.com/buildkite/cli/v3/internal/io"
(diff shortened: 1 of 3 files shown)


pkg/cmd/build/new.go
```diff
@@ -70,33 +70,36 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
...
-		if len(branch) == 0 {
-			p, _, err := f.RestAPIClient.Pipelines.Get(org, pipeline)
-			if err != nil {
-				return err
+	var err error
+	var build *buildkite.Build
+	spinErr := spinner.New().
+		Title(fmt.Sprintf("Starting new build for %s", pipeline)).
+		Action(func() {
+			if len(branch) == 0 {
```
mcncl, line 78: What happens here if the `branch` is set to `   `? Do we trim the white space?

jradtilbrook: Good catch I don't think it was handling that before but I'll add it now

mcncl (approved):

PR #317, opened 2024-07-12
Remove pending command

jradtilbrook:
This is no longer used


jradtilbrook:
This errors because it needs to merge after #316

PR #318, opened 2024-07-12
Unused const removed + clean up

mcncl:
- `OrganizationsSlugConfigKey` is never used except for in its decleration
- Some general formatting such as not having empty lines between decleration and error check


jradtilbrook (approved):
👌

PR #320, opened 2024-07-15
Build resolver should not error when no builds found

lizrabuya:
Update the `cli` such that when no build is resolved, it does not error and return "No Builds found" message instead.

Affected commands:
```
bk build view
bk build rebuild
bk build download
```


jradtilbrook (changes requested):
I don't think this quite hits the target. I think there are still cases where a build not being resolved is an error, but this will never error if no build is found. 

Maybe this should change the case where `bld == nil` and print a message but not error.  
But if `err != nil` then that is still an error.

What this might need to change is the default build resolver. I think it adds a final resolver that will return an error which it should no longer do???

pkg/cmd/build/download.go
```diff
@@ -38,7 +38,8 @@ func NewCmdBuildDownload(f *factory.Factory) *cobra.Command {
 				return err
 			}
 			if bld == nil {
-				return fmt.Errorf("could not resolve a build")
+				fmt.Printf("No build found.\n")
+				return nil
```
jradtilbrook, line 42: I've just realised this is going to cause problems with our writing to stdout. Because this could be run while a spinner is being shown, outputting anything else at the same time causes problems. Not sure what we are supposed to do about that at the moment.

One option is to look at custom error types. So we can return an error of a certain type, and then the consuming function can look at the error and decide whether to "ignore" it

pkg/cmd/build/view.go
```diff
@@ -67,7 +62,7 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 
 			group, _ := errgroup.WithContext(cmd.Context())
 			spinErr := spinner.New().
-				Title("Loading build information").
+				Title("Requesting build information").
```
jradtilbrook, line 65: I prefer to keep "Loading"

pkg/cmd/build/view.go
```diff
@@ -97,6 +92,12 @@ func NewCmdBuildView(f *factory.Factory) *cobra.Command {
 				return err
 			}
 
+			if web {
+				buildUrl := fmt.Sprintf("https://buildkite.com/%s/%s/builds/%d", bld.Organization, bld.Pipeline, bld.BuildNumber)
+				fmt.Printf("Opening %s in your browser\n", buildUrl)
+				return browser.OpenURL(buildUrl)
+			}
+
```
jradtilbrook, line 100: We also don't need to put this below the spinner do we? Once we resolve the build info, we have enough to construct the URL so can open it straight away and return - so its faster for the user.

pkg/cmd/build/download.go
jradtilbrook: Oh actually, this isn't inside another function inside a spinner so it should be okay.

I think the only benefit would be to use `cmd.OurOrStdout` to write to instead

jradtilbrook (changes requested):

pkg/cmd/build/view.go
lizrabuya: The reason for it was to ensure that we only open the page in web if there are no API errors. But I am fine also in putting it back up.

jradtilbrook: Oh yeah but the build is resolved so we should try to open it i think

lizrabuya: I have moved that back up

jradtilbrook: I think you might have not pushed the changes? Its still below

jradtilbrook (approved):

PR #333, opened 2024-08-12
Apply new build resolver to build commands

jradtilbrook:
This refactors the build resolution gear to better support the use of flags to filter the results. All flags now work together to filter the API request down to find a build


internal/build/resolver/options/options.go
```diff
@@ -0,0 +1,88 @@
...
+	if condition {
+		return append(ar, resovler)
+	}
+	return ar
+}
+
+// ResolveBranchFromFlag returns a function that is used to add a branch filter to a build list options
+func ResolveBranchFromFlag(branch string) OptionsResolverFn {
+	return func(context.Context) OptionsFn {
+		return func(options *buildkite.BuildsListOptions) error {
```
mcncl, line 29: I'll get used to this but this feels super hard to glance eval.

mcncl: A function that returns a function that returns a function.

mcncl (dismissed):

internal/build/resolver/options/options.go
jradtilbrook: I've probably done something very wrong

jradtilbrook: Just pushed that change, what do you think now?

internal/build/resolver/options/options.go
```diff
@@ -0,0 +1,75 @@
...
+
+// ResolveBranchFromRepository returns a function that is used to add a branch filter to a build list options
+func ResolveBranchFromRepository(repo *git.Repository) OptionsFn {
+	return func(options *buildkite.BuildsListOptions) error {
+		var branch string
+		if repo != nil && len(options.Branch) == 0 {
+			head, err := repo.Head()
+			if err != nil {
+				return err
+			}
```
jradtilbrook, line 39: Thinking about the case where `bk` is run outside a repository, this would error and not be able to run the command. But perhaps it should just silently not add a branch filter instead. That way its still usable outside a repo if no explicit `--branch` flag is used

lizrabuya (approved):

internal/build/resolver/options/options.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/go-buildkite/v3/buildkite"
+	"github.com/go-git/go-git/v5"
+)
+
+// OptionsResolverFn is a function for adding a branch filter to REST API requests
```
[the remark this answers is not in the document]
jradtilbrook: Yeah thats the idea, it can set any values on the options object passed in

jradtilbrook:
@mcncl what do you think of this PR now? It's simplified slightly now so only double level functions

mcncl (approved):

PR #339, opened 2024-08-21
Add `pipeline view` command

jradtilbrook:
Adds a `pipeline view -w` command for viewing pipeline information.

`-w` will open in a browser instead of showing info in the terminal.

This outputs a few pieces of information about a pipeline that is useful to see. It doesn't show everything, partly because not sure how to display it but also because not sure how important other information is and we can add to it at a later point.

It has some known points of improvement for the future; rendering our custom emoji as emoji instead of the colon syntax and improvements to the colours

Example output:

Screenshot 2024-08-21 at 13 24 32


mcncl:
Has this been tested with a light theme @jradtilbrook?

jradtilbrook:
> Has this been tested with a light theme @jradtilbrook?

No sure hasnt' I was leaning on the majority likely being dark theme and following up with design to improve this dark theme and a light theme at the same time. What do you think @mcncl

lizrabuya (dismissed):

lizrabuya (approved):

PR #340, opened 2024-08-23
Add package push command

moskyb:
**This PR:** Adds a new command to the buildkite cli, `package push`. This command allows users to publish packages to buildkite packages using the CLI.

Attached is the help text for the new command:
```
Push a new package to Buildkite Packages. The package can be passed as a path to a file in the second positional argument,
or via stdin. If passed via stdin, the filename must be provided with the --stdin-file-name flag, as Buildkite
Packages requires a filename for the package.

Usage:
  bk package push registry-name {path/to/file | --stdin-file-name filename -} [flags]

Examples:
$ bk package push my-registry my-package.tar.gz
$ cat my-package.tar.gz | bk package push my-registry --stdin-file-name my-package.tar.gz - # Pass package via stdin, note hyphen as the argument


Flags:
  -h, --help                     help for push
  -n, --stdin-file-name string   The filename to use for the package, if it's passed via stdin. Invalid otherwise.
```

A usage example:
```Bash
$ bkcli package push bennos-gemmos ./test-1.0.0.gem
Created package: test
View it on the web at: http://buildkite.localhost/organizations/buildkite/packages/registries/bennos-gemmos/packages/01918cd0-2568-73aa-8ea3-3d4ec8c0e615
```


pkg/cmd/pkg/package.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdPackage(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:               "package <command>",
+		Short:             "Manage packages",
+		Long:              "Work with Buildkite pipelines.",
```
jradtilbrook, line 13:
```suggestion
		Long:              "Work with Buildkite packages.",
```

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,176 @@
...
+
+func NewCmdPackagePush(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "push --registry <registry> {--file <file> | --file-name <filename> -}",
+		Short: "Push a new package to Buildkite Packages",
+		Example: heredoc.Doc(`
+			$ bk package push --registry my-registry --file my-package.tar.gz
+			$ cat my-package.tar.gz | bk package push --registry my-registry --file-name my-package.tar.gz - # Pass package via stdin, note hyphen as the argument
+		`),
+		Args:      cobra.MaximumNArgs(1),
```
jradtilbrook, line 44: I think no args right?

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,176 @@
...
+				return fmt.Errorf("making create package API request: %w", err)
+			}
+
+			fmt.Println("Created package:")
+			pretty.Println(pkg)
+			return nil
+		},
+	}
+
+	cmd.PersistentFlags().StringP("organization", "o", "", "The organization to create the package in")
```
jradtilbrook, line 93: This isnt how the CLI works. Org is pulled out of config. I'm not sure we want to have a flag for it

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,176 @@
...
+	cmd.PersistentFlags().StringP("registry", "r", "", "The slug for the registry to create the package in")
+	cmd.PersistentFlags().StringP("file", "f", "", "The path to the package file to upload. Cannot be used when package is passed via stdin")
+	cmd.PersistentFlags().StringP("file-name", "n", "", "The filename to use for the package, if it's passed via stdin. Invalid otherwise.")
```
jradtilbrook, line 96: I dont think we want persistent flags here? arent those types passed to child commands?

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,176 @@
...
+		Use:   "push --registry <registry> {--file <file> | --file-name <filename> -}",
+		Short: "Push a new package to Buildkite Packages",
+		Example: heredoc.Doc(`
+			$ bk package push --registry my-registry --file my-package.tar.gz
+			$ cat my-package.tar.gz | bk package push --registry my-registry --file-name my-package.tar.gz - # Pass package via stdin, note hyphen as the argument
+		`),
+		Args:      cobra.MaximumNArgs(1),
+		ValidArgs: []string{"-"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg, err := populateFlags[newPackageConfig](cmd.PersistentFlags())
```
jradtilbrook, line 47: i have a feeling the way flags are handled by other commands is much simpler - just giving you a primitive type to work with

jradtilbrook (commented):
This is very cool to see packages making its way into the CLI 💪 

I will defer to you with more go experience, but have noted below lots of different ways than what we currently have for other commands if you can comment on those

pkg/cmd/pkg/package.go
moskyb: 😬

pkg/cmd/pkg/push.go
moskyb: oh shoot you're totally right! this flag isn't even used, it's from a previous iteration from before i realised that org is stored. i'll remove this line

pkg/cmd/pkg/package.go
jradtilbrook: 😜

pkg/cmd/pkg/push.go
moskyb: oh yup, you're totally right - will make these regular flags

pkg/cmd/pkg/push.go
moskyb: ah, the `-` at the end, specifying that the command should receive input from stdin, is technically an arg

pkg/cmd/pkg/push.go
moskyb: yeah, i originally handled it that way but got frustrated with pointer dereferencing and nil checking, so decided to replace it with the much more consistent and easy to read logic of (checks notes) reflection.

in retrospect, i've probably overcomplicated things here - happy to switch back to the other way ;)

pkg/cmd/pkg/push.go
jradtilbrook: oh yeah okay thats tricky! 👌

pkg/cmd/pkg/push.go
jradtilbrook: Haha nice. I dont think we've used actual pointer variables but just passed them into the flags by reference. Not sure if that works for this though 🤷‍♂️

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,176 @@
...
+		return false, fmt.Errorf("failed to stat stdin: %w", err)
+	}
+
+	readable := (stat.Mode() & os.ModeCharDevice) == 0
+	return readable, nil
+}
+
+func NewCmdPackagePush(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "push --registry <registry> {--file <file> | --file-name <filename> -}",
```
jradtilbrook, line 38: One other thing we've tried to do is allow each command to do "something" useful by default without any flags. And so all flags are optional to modify the default behaviour.

If there is something that is required, I think it should be an argument instead? Ie. should registry (or maybe both flags) be positional arguments instead?

go.mod
```diff
@@ -24,6 +24,8 @@ require (
 	golang.org/x/sync v0.7.0
 )
 
+replace github.com/buildkite/go-buildkite/v3 => ../go-buildkite
```
jradtilbrook, line 27: probably need to do another release and reverse this?

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,152 @@
...
+				from = cmd.InOrStdin()
+			}
+
+			r, w := io.Pipe()
+			go func() {
+				if _, err := io.Copy(w, from); err != nil {
+					w.CloseWithError(fmt.Errorf("failed to read package from input: %w", err))
+				}
+				w.Close()
+			}()
```
jradtilbrook, line 67: I'm trying to wrap my head around this part. Theres a pipe so we can read from one and write to another but have another reader as well right? I'm not sure why its needed on top of just the file or stdin

jradtilbrook (approved):

pkg/cmd/pkg/push.go
moskyb: note for future travellers: we talked about this offline, and i realised that this actually does nothing and just makes things complicated for no reason. i had thought that this avoided storing the contents of `from` in-memory, but just passing `from` directly into the call to `CreatePackage` also doesn't store anything in memory, so we don't need this malarkey.

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,159 @@
...
+				from = file
+			case cfg.StdinFileName != "":
+				packageName = cfg.StdinFileName
+				from = cmd.InOrStdin()
+			}
+
+			pkg, _, err := f.RestAPIClient.PackagesService.Create(f.Config.OrganizationSlug(), cfg.RegistrySlug, buildkite.CreatePackageInput{
+				Filename: packageName,
+				Package:  from,
+			})
```
mcncl, line 64: Should we add a check to ensure this isn't a nil pointer?

pkg/cmd/pkg/push.go
```diff
@@ -0,0 +1,159 @@
...
+			switch {
+			case cfg.FilePath != "":
+				packageName = cfg.FilePath
+
+				file, err := os.Open(cfg.FilePath)
+				if err != nil {
+					return fmt.Errorf("couldn't open file %s: %w", cfg.FilePath, err)
+				}
+
+				defer file.Close()
```
mcncl, line 53: Should this come before the error check to ensure the file is closed even in the result of an error?

pkg/cmd/pkg/push.go
moskyb: yeah, i was wondering about that earlier... theoretically `loadAndValidateConfig` should ensure that one or the other of the two sources is selected, but it can't hurt to double check.

given that neither stdin nor a file being available is likely an error in programming, i add a 
```
default:
  panic("oh no!")
```
to the switch block above.

pkg/cmd/pkg/push.go
moskyb: ooh yup, good call

DrJosh9000: If `os.Open` returns an error, then `file` is (probably) nil, so `defer file.Close()` before checking `err` would probably cause a nil dereference panic.

cmd/bk/main.go
```diff
@@ -30,7 +30,7 @@ func mainRun() int {
 	}
 
 	if err := rootCmd.ExecuteContext(ctx); err != nil {
-		fmt.Fprintf(os.Stderr, "failed to execute command: %s\n", err)
+		// Errors already get printed by the cobra command runner
```
jradtilbrook, line 33: Oh my god. That has been bugging me but I keep forgetting to create a ticket/look into it. Thank you!!

pkg/cmd/pkg/push.go
jradtilbrook: Is this upload where the bulk of the time will be spent? Ie if its uploading a few MB or many more, it could take a while. We have a spinner package we use to give the user some feedback at least. I made a patch you can apply if you like: https://gist.github.com/jradtilbrook/a3af75bab23d26edb6f2bb2291bf17d1

I'm assuming simply opening the file won't take long and not worth including in the spinner

PriyaSudip (approved):

mcncl (approved):

PR #343, opened 2024-08-29
API command

mcncl:
## Changes
This adds an `api` command to the CLI, which can be used to make any API call that can currently be made via REST in the Buildkite [REST API](https://buildkite.com/docs/apis/rest-api). 

Non-existing Pipeline/resource:
CleanShot 2024-08-29 at 20 59 57@2x

Existing Pipeline:
CleanShot 2024-08-29 at 21 00 48@2x

With data/creating a Pipeline:
CleanShot 2024-08-30 at 06 52 21

### DIFF
--- a/pkg/cmd/api/api.go
+++ b/pkg/cmd/api/api.go
@@ -47,6 +47,9 @@ func NewCmdAPI(f *factory.Factory) *cobra.Command {
         "name": "My Updated Cluster",
       }
       '
+
+      # To get all test suites
+      $ bk api --analytics /suites
       `),
 		PersistentPreRunE: validation.CheckValidConfiguration(f.Config),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/pkg/cmd/api/api.go
+++ b/pkg/cmd/api/api.go
@@ -1,0 +1,122 @@
+package api
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	"github.com/spf13/cobra"
+)
+
+var (
+	method      string
+	headers     []string
+	data        string
+	baseURL     string
+	graphqlFlag bool
+	queryFile   string
+)
+
+func NewCmdAPI(f *factory.Factory) *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "api <endpoint>",
+		Short: "Interact with the Buildkite API",
+		Long:  "Interact with either the REST or GraphQL Buildkite APIs",
+		Args:  cobra.MaximumNArgs(1),
+		Example: heredoc.Doc(`
+      # To make an API call
+      $ bk api pipelines/example-pipeline/builds/420
+      `),
+		PersistentPreRunE: validation.CheckValidConfiguration(f.Config),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return apiCaller(cmd, args, f)
+		},
+	}
+
+	cmd.Flags().StringVarP(&method, "method", "X", "GET", "HTTP method to use")
+	cmd.Flags().StringArrayVarP(&headers, "header", "H", []string{}, "Headers to include in the request")
+	cmd.Flags().StringVarP(&data, "data", "d", "", "Data to send in the request body")
+	cmd.Flags().StringVar(&baseURL, "base-url", fmt.Sprintf("https://api.buildkite.com/v2/organizations/%s", f.Config.OrganizationSlug()), "Base URL for the API")
+	cmd.Flags().BoolVar(&graphqlFlag, "graphql", false, "Use GraphQL API")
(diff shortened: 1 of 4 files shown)


pkg/cmd/api/api.go
```diff
@@ -0,0 +1,122 @@
...
+	cmd.Flags().StringVarP(&queryFile, "file", "f", "", "File containing GraphQL query")
+
+	return &cmd
+}
+
+func apiCaller(cmd *cobra.Command, args []string, f *factory.Factory) error {
+	var endpoint string
+
+	if len(args) > 1 {
+		return fmt.Errorf("Incorrect numebr of arguments. Expected 1, got %d", len(args))
```
jradtilbrook, line 57:
```suggestion
		return fmt.Errorf("Incorrect number of arguments. Expected 1, got %d", len(args))
```

pkg/cmd/api/api.go
```diff
@@ -0,0 +1,122 @@
...
+		endpoint = args[0]
+	}
+	url := baseURL + endpoint
+
+	var req *http.Request
+	var err error
+
+	if graphqlFlag {
+		return errors.New("GraphQL not supported at this time.")
+	}
```
jradtilbrook, line 72: Maybe leave it off for now then and create a linear?

jradtilbrook (commented):
Freakin awesome 👌

pkg/cmd/api/api.go
```diff
@@ -0,0 +1,122 @@
...
+	cmd.Flags().StringArrayVarP(&headers, "header", "H", []string{}, "Headers to include in the request")
+	cmd.Flags().StringVarP(&data, "data", "d", "", "Data to send in the request body")
+	cmd.Flags().StringVar(&baseURL, "base-url", fmt.Sprintf("https://api.buildkite.com/v2/organizations/%s", f.Config.OrganizationSlug()), "Base URL for the API")
+	cmd.Flags().BoolVar(&graphqlFlag, "graphql", false, "Use GraphQL API")
+	cmd.Flags().StringVarP(&queryFile, "file", "f", "", "File containing GraphQL query")
+
+	return &cmd
+}
+
+func apiCaller(cmd *cobra.Command, args []string, f *factory.Factory) error {
```
lizrabuya, line 53: I know that this is just an initial work, but just wondering if we are going to move this to the same resolver models that was started for the other cli commands?

pkg/cmd/api/api.go
```diff
@@ -0,0 +1,122 @@
...
+		PersistentPreRunE: validation.CheckValidConfiguration(f.Config),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return apiCaller(cmd, args, f)
+		},
+	}
+
+	cmd.Flags().StringVarP(&method, "method", "X", "GET", "HTTP method to use")
+	cmd.Flags().StringArrayVarP(&headers, "header", "H", []string{}, "Headers to include in the request")
+	cmd.Flags().StringVarP(&data, "data", "d", "", "Data to send in the request body")
+	cmd.Flags().StringVar(&baseURL, "base-url", fmt.Sprintf("https://api.buildkite.com/v2/organizations/%s", f.Config.OrganizationSlug()), "Base URL for the API")
```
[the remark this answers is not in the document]
mcncl: Ha, yeah, I was just getting it _working_ first. Will do some cleanup pre-marking as ready.

pkg/cmd/api/api.go
mcncl: @lizrabuya IMO, which I'm happy to have changed, the intent of the `api` command is to return API responses too, the API would return a JSON blob and so that should be the value returned by this command.

mcncl: I don't see a use for the build resolver things in this part of the CLI

jradtilbrook: I think allowing things to be filled in from the current repo would be nice, but not necessary at this point. Ie `/pipelines/:pipeline/builds` or something which would resolve the current directories repo/pipeline

lizrabuya: yeah, maybe we visit that later. It's too early to decide right now. Thanks both!

PriyaSudip (approved):

james2791 (dismissed):
:shipit:

lizrabuya (approved):

pkg/cmd/api/api.go
```diff
@@ -0,0 +1,142 @@
...
+		Use:   "api <endpoint>",
+		Short: "Interact with the Buildkite API",
+		Long:  "Interact with either the REST or GraphQL Buildkite APIs",
+		Args:  cobra.MaximumNArgs(1),
+		Example: heredoc.Doc(`
+      # To get a build
+      $ bk api /pipelines/example-pipeline/builds/420
+
+      # To create a pipeline
+      $ bk api --method POST /pipelines --data '
```
jradtilbrook, line 36: It would even be cool if we automatically used POST when data is included since its not valid for GET anyway. And then people can specify PUT/PATCH when necessary to change that. A thing for another PR though

mcncl: Hmm, the problem is you also pass `data` to a PUT, so setting it to POST as default might cause confusion

jradtilbrook (approved):

james2791 (approved):

jradtilbrook (approved):

PR #349, opened 2024-09-06
Remove AI command

mcncl:
## Changes
We don't need an AI command for now, we aim to maybe support this in the future but given the complexities of AI/tokens/data there are other things we can work on which would have a bigger impact.


lizrabuya (approved):

jradtilbrook (approved):
Sure you want to remove the actual code? Not just the addition to root.go?

mcncl:
It'll be in the commit history if we need it 🙂

PR #350, opened 2024-09-09
Show spinner while pushing package

jradtilbrook:
This modifies `package push` to show a spinner while its uploading so there is some user feedback.

Also adds a `pkg` alias for `package`.

### DIFF
--- a/pkg/cmd/pkg/package.go
+++ b/pkg/cmd/pkg/package.go
@@ -9,6 +9,7 @@ import (
 func NewCmdPackage(f *factory.Factory) *cobra.Command {
 	cmd := cobra.Command{
 		Use:               "package <command>",
+		Aliases:           []string{"pkg"},
 		Short:             "Manage packages",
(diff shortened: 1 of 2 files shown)


mcncl:
Can you add something to show the process? Like a gif so we can see what happens when the command runs, when it's running and when it's finished?

jradtilbrook:
> Can you add something to show the process? Like a gif so we can see what happens when the command runs, when it's running and when it's finished?

@mcncl I started work on a progress bar in https://gist.github.com/jradtilbrook/4691ab924bc89743e7e4b6d8a3424173 but this approach has quite a few issues; it only shows based on reading the file off disk, not actually uploading it, and even after we do that, the server takes some time to open and inspect it to save it so we are left with a chunk of time we cannot know much about.

There is going to be improvements to the `push` command to use multipart uploads so I'm think we wait for that, which should give us a much more accurate progress measure

mcncl (approved):

PR #354, opened 2024-09-10
Add user agent for http requests

jradtilbrook:
Add user-agent header for all api requests. This will allow us to check usage over time and ensure we can upgrade without breaking usage


mcncl (approved):

jradtilbrook:
This sets a UA like: `go-buildkite/3.13.0 buildkite-cli/DEV` which I think will suffice. We could include go version in that but I dont see that being a big cause of problems that we'd like to track, given we control the compilation and go version anyway

PR #365, opened 2024-09-18
Remove uses of AI/OpenAI configs for now

mcncl:
## Changes
We aren't using AI in the CLI for now, so let's remove any use of external packages that _are_ AI related as well as the internal configs for those.

- Removes AI token setter/getter
- Removes check for OpenAI token in config file
- Removes use of external package


moskyb (approved):

lizrabuya (dismissed):
👍

PriyaSudip (approved):

PR #367, opened 2024-09-19
SUP-2587 | Support inviting users via the CLI

mcncl:
## Changes
We can support user invites via the CLI.

- GraphQL query in new `organization` internal dir to get the organization ID which is required in some queries/mutations
- User central command
- Invite command which accepts 1 or many emails
    - There is internal validation on emails that will return the error if an incorrect/no email is used

### Caveats

- Doesn't currently apply roles/teams. We can add this in the future


james2791 (approved):
LGTM ✅ 

(Will need a codeowner)

lizrabuya (approved):
👍

PR #368, opened 2024-09-20
support flag for folks who use that kind of thing

mcncl:
## Changes

Some of our older users 😛 will be used to `--version` or `-v` for printing the version, so let's support that.


catkins (dismissed):

lizrabuya (approved):

sj26:
Nice!

https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html

> All programs should support two standard options: ‘--version’ and ‘--help’.

sj26:
Note that `-V` (upper case) is typically for "version", and `-v` (lower case) is typically `--verbose`.

PR #377, opened 2024-10-09
Support using a file for env

mcncl:
## Changes

We implemented support for `-e` flags at `build new` time, but some folks have a lot of environment variables they wish to set.

- Support using a file, `-f` for the environment

I'm open to suggestions if this flag/short version could be better


lizrabuya (approved):
👍

Issue #378, opened 2024-10-09
Allow setting `ignore_pipeline_branch_filters` in `bk build new`


amartani:
Creating a build using `bk build new` in a pipeline that has GitHub integration disabled fails with:

```
bk build new --pipeline=$org_name/$pipeline_name
Error: POST https://api.buildkite.com/v2/organizations/$org_name/pipelines/$pipeline_name/builds: 422 Branches have been disabled for this pipeline
```

To create builds for this pipeline via the API, we need to set `ignore_pipeline_branch_filters: true` in the POST call. Could this option be added as a flag for the CLI?

Thank you!

mcncl:
@amartani thanks for raising the issue. I’d suggest using the `api` command for this use case at the moment.

raylu:
this is a rather serious regression that break `bk build` (our main use for the CLI) for most of our pipelines

the pipelines we are most likely to want to trigger manually (via `bk build new`) are the ones most likely to not have the GH integration and be affected by this bug

mcncl:
@raylu @amartani I've opened https://github.com/buildkite/cli/pull/379 to fix this

mcncl:
@amartani @raylu this feature is included in `3.2.0`. Minor bump as there are other additions in there too; supporting `-f` for passing in an environment file, etc

PR #379, opened 2024-10-10
Support branch filter ignoring

mcncl:
## Changes
Closes https://github.com/buildkite/cli/issues/378

- Allows the use of `-i` or `--ignore-branch-filters` when creating a new build so that filter rules can be bypassed


lizrabuya (approved):
👍

PR #380, opened 2024-10-21
Update go-buildkite to v4.0.0

moskyb:
This PR updates [go-buildkite](https://github.com/buildkite/go-buildkite) to v4! It deals with the major breaking changes:
- strips out the many pointer derefs we had, because they're no longer needed!
- passes contexts to all of the calls to go-buildkite calls
  - there's a couple in the builds resolver that i've just plugged `context.TODO()` into, because the build resolver scares me and i'm afraid to touch it


sj26:
This looks pretty fine to me, but looks like a red build at the moment :eyes:

sj26 (approved):
+1 in principle, when green

PriyaSudip (approved):

PR #401, opened 2024-11-19
Use standardised flag; not camel case

mcncl:
## Changes

- Standardise the use of the envFile flag to that it is used with --env-file instead
- Add a deprecated warning to the envFile flag
- If envFile is used then normalise to --env-file


PriyaSudip (dismissed):

lizrabuya (approved):
LGTM!

PR #406, opened 2024-11-26
Allow for an alternate buildkite.com to be specified

CerealBoy:
Given a `BUILDKITE_GRAPHQL_ENDPOINT` env is found, use it as the base for the GraphQL API calls being made instead of the built-in default.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -121,6 +121,14 @@ func (conf *Config) ConfiguredOrganizations() []string {
 	return keys
 }
 
+func (conf *Config) GetGraphQLEndpoint() string {
+	value := os.Getenv("BUILDKITE_GRAPHQL_ENDPOINT")
+	if value != "" {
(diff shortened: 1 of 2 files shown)


moskyb (approved):
LGTM!

edit, looks like i don't have approve power 😅

dabarrell:
Looks like this only handles graphql - it'd be good if we didn't have to specify the graphql and rest API base URLs separately

mcncl:
> LGTM!
> 
> edit, looks like i don't have approve power 😅

You can approve, but it means nothing 😛

mcncl (approved):

CerealBoy:
> Looks like this only handles graphql - it'd be good if we didn't have to specify the graphql and rest API base URLs separately

@dabarrell There doesn't appear to be any rest API usage at the moment, only graphql.

PR #407, opened 2024-12-02
Allow non-interactive `bk configure` command

stone-w4tch3r:
Currently `bk configure` is interactive and this prevents from using it in poduction, eg github actions. This PR adds `--org` and `--token` options for non-interactive bk CLI configure.

**Note**: due to multiple bugs in current buildkite packages system, I can't verify that I really can push packages after non-interactive configuration. But I can't push packages with current release version too. I am not familiar with other buildkite features, so unfortunately can't test my changes


mcncl:
Thanks for the contribution @stone-w4tch3r! ❤️

PR #408, opened 2024-12-02
Environment Variable config

moskyb:
**This PR:** Allows users to specify their buildkite API tokens and organization slugs via environment variable, allowing them to end-run around the `bk configure` command. This should make running the `bk` cli in headless environments (like CI jobs 👀) much easier.

Environment variables take precedence over config items specified in files, as is fairly standard across CLI tools.

Example:
```
$ rm ~/.config/bk.yaml
$ export BUILDKITE_API_TOKEN=bkua_haha_nice_try 
$ export BUILDKITE_ORGANIZATION_SLUG=bennos-extra-secret-org 
$ bk pipeline view "my-cool-pipeline"
  My Cool Pipeline: Possibly the coolest pipeline around
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

                         Speed: 11s    Reliability: 80%

    agents:
      queue: default

    steps:
      - command: buildkite-agent pipeline upload
```

Additionally, it adds another envar for `BUILDKITE_REST_API_HOST`, similar to the graphql host envar added in #406. This should make local development easier.

🍷🧀 This PR should pair with #407 nicely - they allow gradations in how interactive you want CLI configuration to be

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -15,6 +15,7 @@ import (
 	"slices"
 
 	"github.com/buildkite/cli/v3/internal/pipeline"
+	"github.com/buildkite/go-buildkite/v4"
 	"github.com/go-git/go-git/v5"
 	"github.com/spf13/afero"
 	"github.com/spf13/viper"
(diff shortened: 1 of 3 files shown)


internal/config/config.go
```diff
@@ -169,16 +184,14 @@ func (conf *Config) SetPreferredPipelines(pipelines []pipeline.Pipeline) error {
...
 }
 
-func firstNonEmpty[T comparable](t ...T) T {
-	var empty T
-
-	for _, k := range t {
-		if k != empty {
+func firstNonEmpty(s ...string) string {
+	for _, k := range s {
+		if k != "" {
```
mcncl, line 189: Was there a reason for changing this? I think we only ever give it strings, but in its current form it could be used with any comparable type.

moskyb: nothing other than YAGNI - i figured if we weren't using generics we probably shouldn't use generics, but i'm more than happy to change it back if you like

mcncl: Nah, not at all, think it was in as a just in case, but we can re-visit if it's ever needed

mcncl (approved):

PR #409, opened 2024-12-05
Only use a spinner on a tty

sj26:
I tried to use bk to push a package from GitHub Actions, but it failed:

https://github.com/sj26/github-actions-to-buildkite-packages-javascript-example/actions/runs/12133880719/job/33830150889#step:5:8

Screenshot 2024-12-05 at 11 52 33 AM

Pretty sure that message comes via `spinErr`. I think it's because it's trying to start spinner, and that seems to require a tty.

This wraps up the spinner a little function that checks if we're in a tty when starting a spinner, and skips the spinner if we're not.

I haven't attempted tests yet .. I just wanted to see if this works.

I did test this locally and could still push a package and it showed a spinner. I couldn't figure out quickly how to simulate a no-tty environment.

### DIFF
--- a/internal/io/spinner.go
+++ b/internal/io/spinner.go
@@ -1,0 +1,21 @@
+package io
+
+import (
+	"os"
+
+	"github.com/charmbracelet/huh/spinner"
+	"github.com/mattn/go-isatty"
+)
+
+func SpinWhile(name string, action func()) error {
+	if isatty.IsTerminal(os.Stdout.Fd()) {
+		action()
+
+		return nil
+	}
+
+	return spinner.New().
+		Title(name).
+		Action(action).
+		Run()
+}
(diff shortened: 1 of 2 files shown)


mcncl (approved):

sj26:
Sadly this doesn't seem to have fixed my issue 😢

internal/io/spinner.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"os"
+
+	"github.com/charmbracelet/huh/spinner"
+	"github.com/mattn/go-isatty"
+)
+
+func SpinWhile(name string, action func()) error {
+	if isatty.IsTerminal(os.Stdout.Fd()) {
```
sj26, line 11: This should have been:


```suggestion
	if !isatty.IsTerminal(os.Stdout.Fd()) {
```

PR #414, opened 2024-12-11
initial api scoping checks

mcncl:
## Changes

I think we can do a better job around API permissions; currently we just try to perform a command even if the API token doesn't have scope.

- Initial framework for possible API scopes
- Helper commands so that scopes can be checked using the `scopes` package
- A getter for API scopes which syncs/caches
- Usage of scope checker in `build/new.go`


lizrabuya (approved):
Looks good! 🚀

PR #420, opened 2024-12-15
api scopes for all build commands

mcncl:
## Changes

- Introduce a pre-run check for API scope permissions
- Error where permissions not included with token


lizrabuya (commented):
@mcncl , everything looks good! But is it possible to move the command scope validation into the validation package since it will be reused across different commands?

pkg/cmd/build/cancel.go
```diff
@@ -28,6 +29,23 @@ func NewCmdBuildCancel(f *factory.Factory) *cobra.Command {
 		Long: heredoc.Doc(`
 			Cancel the given build.
 		`),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
lizrabuya, line 32: Request to optimise preRun function that appears on several portions in code

lizrabuya (changes requested):
Looks good!

pkg/cmd/build/cancel.go
mcncl: Sure! How would you optimise this code? I'm not overly certain on a way to reduce it further

lizrabuya: Oh, the code  to check the scope is fine. But maybe move the implementation in one place (like in pkg/cmd/build/build.go) and refer to it in the other subcommands ?

lizrabuya: hi @mcncl , we can implement this feedback in an enhancement.

lizrabuya (approved):
👍

PR #426, opened 2024-12-30
SUP-2875 | produce actionable error on missing default branch

mcncl:
## Changes

We can check the returned value of the pipeline targeted as it's a field on the struct

```go
if p.DefaultBranch == "" {
	actionErr = fmt.Errorf("No default branch set for pipeline %s. Please specify a branch using --branch (-b)", pipeline)
	return
}
```

Here we just check that the `default_branch` of the pipeline isn't null, if it is then we produce a helpful error to use `--branch` in the command, else the build will be created on the _default branch_ for the pipeline.`

### DIFF
--- a/pkg/cmd/build/new.go
+++ b/pkg/cmd/build/new.go
@@ -143,6 +143,12 @@ func newBuild(ctx context.Context, org string, pipeline string, f *factory.Facto
 					actionErr = fmt.Errorf("Error getting pipeline: %w", err)
 					return
 				}
+
+				// Check if the pipeline has a default branch set
+				if p.DefaultBranch == "" {
+					actionErr = fmt.Errorf("No default branch set for pipeline %s. Please specify a branch using --branch (-b)", pipeline)
+					return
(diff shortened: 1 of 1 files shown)


pkg/cmd/build/new.go
```diff
@@ -143,6 +143,12 @@ func newBuild(ctx context.Context, org string, pipeline string, f *factory.Facto
 					actionErr = fmt.Errorf("Error getting pipeline: %w", err)
 					return
 				}
+
+				// Check if the pipeline has a default branch set
+				if p.DefaultBranch == "" {
+					actionErr = fmt.Errorf("No default branch set for pipeline %s. Please specify a branch using --branch (-b)", pipeline)
+					return
+				}
```
mcncl, line 151: This is Go's way of checking that there is _some_ value to a string, we don't care what that value is (`foo` is as valid as `main` for a default branch)

PriyaSudip (approved):

PR #433, opened 2025-01-15
Implement `bk artifacts download`

lizrabuya:
The `bk artifacts download` can be used to download a specific artifact by its UUID. 

```
Usage:
  bk artifacts download <artifact UUID>

Examples: 
$ bk artifacts download 0191727d-b5ce-4576-b37d-477ae0ca830c 

```


internal/artifact/genqlient.yaml
mcncl, line 1: Do we need this file? It already exists at the root of the project and targets `internal/**/*.graphql`

lizrabuya: Must have wrongly checked it in. Will remove it

mcncl (approved):

PR #441, opened 2025-02-13
🎨 Improve the structure of some internal packages

mcncl:
## Changes
- We have moved `scopes` to our `validation` package
- We have split our `validation` package out for clarity/easy addition of rules
- We **test** the validation works
- Because of this, we needed to adjust our `build` internals
- Adds a `view` package to our `/internal/build`, which contains the individual focuses of the `build` package


internal/build/view/view.go
```diff
@@ -0,0 +1,79 @@
...
+	BuildNumber  int
+	Web          bool
+}
+
+func (o *ViewOptions) Validate() error {
+	v := validation.New()
+	v.AddRule("Organization", validation.Required)
+	v.AddRule("Organization", validation.Slug)
+	v.AddRule("Pipeline", validation.Required)
+	v.AddRule("Pipeline", validation.Slug)
```
mcncl, line 25: This is an example of how we can add validation to commands

internal/build/view/view.go
```diff
@@ -0,0 +1,79 @@
...
+		for _, j := range v.Build.Jobs {
+			if j.Type == "script" {
+				sections = append(sections, shared.RenderJobSummary(j))
+			}
+		}
+	}
+
+	// Add artifacts if present
+	if len(v.Artifacts) > 0 {
+		sections = append(sections, v.renderArtifacts())
```
mcncl, line 70: Our rendering functions are now split out in to their own files, so this can be cleaner and we can just add more if needed

internal/validation/rule.go
```diff
@@ -0,0 +1,5 @@
+package validation
+
+type Rule interface {
+	Validate(value interface{}) error
+}
```
mcncl, line 5: A `Rule` just needs to have a function, `Validate` which takes in a `value` of `any` type and returns an error

PriyaSudip (commented):
Looks good, could you fix the failing pipeline so we can go ahead

PriyaSudip (approved):

PR #442, opened 2025-02-13
✨ Show org in prompt

mcncl:
## Changes
- Adds a prompt configurator so that the current org can show in the prompt when in a project dir
- Adds a README on using shell prompt integration
- Updates base README to link to new doc

CleanShot 2025-02-14 at 07 16 30


lizrabuya (approved):
Nice! 🚀

PR #451, opened 2025-02-27
🐛 Fix the API req for validate

mcncl:
# Fix validate command; don't require API token

The `validate` command was still inheriting the `pipeline.go` requirement for config check. This meant that API keys were still needed to validate a pipeline.

## Changes

This adjusts the way we validate in our `validate/config.go`; it allows us to pass commands which do not require API tokens

- `pkg/cmd/validation` checks on the path of a command via `getCommandPath`, if that path is in `CommandsNotRequiringToken` then we don't need an API token

## Improvements

This should enable us to set other commands that don't require API tokens


lizrabuya (approved):
LGTM!

PR #455, opened 2025-03-07
Add a testutil package

mcncl:
# Implement a `testutil` package

Currently we replicate setup code across a few packages/commands, eg.

- internal/pipeline/resolver/repository_test.go
- internal/pipeline/resolver/cli_test.go 
- internal/agent/stoppable_test.go
- pkg/cmd/pkg/push_test.go

These all use some kind of generic set up, but they don't need to.

## Changes

This PR introduces a `testutil` package to our `/internal` directory, for use with tests.

We can use this in both `/internal` and `/pkg` files (note the `pkg/push_test.go` changes).

## Documentation

In the interest of making contributions to the CLI easier, we've added a `README` to this package to guide folks on how to best use it when creating tests, or when creating new utility files that we can use elsewhere.

## Testing

Our standard testing commands should still pass, functionality has been moved and improved, we shouldn't see any failures.

```bash
go test ./... -v
```

### DIFF
--- a/internal/agent/stoppable_test.go
+++ b/internal/agent/stoppable_test.go
@@ -1,12 +1,11 @@
 package agent
 
 import (
-	"bytes"
 	"errors"
-	"io"
 	"testing"
 	"time"
 
+	"github.com/buildkite/cli/v3/internal/testutil"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/x/exp/teatest"
 )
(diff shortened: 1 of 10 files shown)


DrJosh9000 (commented):
:car: beep beep it's your friendly drive-by Go reviewer!

I feel behooved to point out:

* Implementing a custom assertion library means someone reading this code not only has to learn how Go testing works in general, but the specific "language" of the custom assertion library
  * ("teatest" is not immune from this criticism)
* `testutil.AssertTeaOutput` is OK as naming goes, it's somewhat descriptive. The implementation is a step above some of the code it replaces ("Output did not match" is far worse than "Expected output to contain %q, got %q" and you earn 1 point for using `%q`) but still lacks info that would shorten the test/fix loop (you could put more helpful info in the logs to avoid having to dig into the test implementation, basically):
  * which function or type+method was called to produce the wrong output
  * what args were passed to the function/method that was called to produce the wrong output

PR #457, opened 2025-03-18
Fix build download command required scopes

Inozuma:
# Description

Hello, as I was trying to download the logs from one of my pipeline build, I encountered the following error:

```
❯ bk build download --pipeline my_pipeline build_number
Error: missing required scopes: read_buildsread_artifactsread_build_logs
```

After checking my token scopes and digging around, I found that the `"requiredScopes"` annotation of the `build download` command was missing the `,` separator.

As the fix was straightforward (and successfully downloaded my logs with this fix), I did not open any issue.
Let me know if you want me to add anything.

### DIFF
--- a/internal/validation/scopes/scopes.go
+++ b/internal/validation/scopes/scopes.go
@@ -2,13 +2,40 @@ package scopes
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
 )
 
+// Scopes is a helper class to manipulate a list of scopes.
+type Scopes []Scope
+
+// NewScopes creates a new Scopes from the given list of scopes.
+func NewScopes(scopes ...Scope) Scopes {
+	return scopes
+}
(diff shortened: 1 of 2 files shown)


mcncl:
Thanks for the PR, @Inozuma! While it would fix this issue, I wonder should we implement a helper function to take in a list of scopes and return a string, given the solution in this PR would need iterating on if the permissions ever changed/increased.

Inozuma:
Yes, indeed. What do you think of a new type `type Scopes []Scope` with a `String` method?

It could also be useful for additional methods when manipulating multiple `Scope`.

mcncl:
@Inozuma absolutely, sounds great. You're welcome to open a PR, or create a request via Issues.

mcncl:
@Inozuma Thanks! I'll have a look at this in the morning and see if we can merge it in

mcncl (approved):

PR #476, opened 2025-04-29
[SUP-2184] Only write local config in git repo

jykingston:
**This is not ready for merge!**

Running the BK CLI tool from any directory will create a local config. This PR is to change this functionality so that it does not create the local config unless the directory is a valid git repo.


pkg/cmd/configure/add/add.go
```diff
@@ -23,9 +23,10 @@ func NewCmdAdd(f *factory.Factory) *cobra.Command {
 }
 
 func ConfigureWithCredentials(f *factory.Factory, org, token string) error {
-	if err := f.Config.SelectOrganization(org); err != nil {
+	if err := f.Config.SelectOrganization(org, f.GitRepository == nil); err != nil {
```
mcncl, line 26: This should be `!=`, rather than `==`; we want to check that the git repository is not `nil`, rather than it being `nil`

mcncl (commented):
I think this will also cause issues on `configure`, where an API token will be required and not allow the user to `configure` with an incorrect one, but an API token should not be needed for certain commands, like `bk pipeline validate`.

PriyaSudip (commented):
Changes looks good, will wait for Test and Lint to go green.

PriyaSudip (approved):

PR #478, opened 2025-05-18
Add documentation generation with llms.txt support

toolmantim:
Adds a new 'docs' command that can generate documentation in Markdown format and in llms.txt format for better AI assistant support.

This PR also adds a link to the LLM documentation file in the README.md, making it easier for users and AI assistants to discover and access the comprehensive CLI documentation.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -8,6 +8,13 @@ release:
   prerelease: auto
   make_latest: false
   mode: replace
+  extra_files:
+    - glob: ./llms.txt
+
+before:
+  hooks:
+    # Generate llms.txt file before creating the release
+    - go run ./cmd/bk docs --format llms --output ./llms.txt
 
 changelog:
   use: github
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ Want to display your current Buildkite organization in your shell prompt? Check 
 
 We'd love to hear any feedback and questions you might have. Please [file an issue on GitHub](https://github.com/buildkite/cli/issues) and let us know!
 
+## LLM Documentation
+
+We provide a comprehensive [LLM-optimized documentation file](https://github.com/buildkite/cli/releases/latest/download/llms.txt) for AI language models to better understand the CLI.
+
 ## 🔨 Development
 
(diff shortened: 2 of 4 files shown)


toolmantim:
Experimenting a little more with Claude, it does attempt to guess the args and then run `help` if it gets it wrong:

CleanShot 2025-05-18 at 16 48 00@2x

But then `help` doesn't give you full docs for sub-commands too.

There's probably a two pronged approach:

1. Teach the foundation models the complete docs of the `bk` cli command (which this PR is meant to help with, as well as letting users paste in the llms URL into an agent chat thread)
2. Helping the LLM agent figure out the right command based on a `bk` binary invocation (either expanding the default `bk help` output, or adding a new `bk help --llms` arg that's mentioned in `bk help`)

toolmantim:
I'll move this back to draft, and would love your input on the above.

toolmantim:
It appears it should also be llms-full.txt for a complete doc

mcncl:
👋 Appreciate the ongoing contribution!

> But then help doesn't give you full docs for sub-commands too.

Yeah, I think that's a caveat of using Cobra; something like Kong would allow us to get that documentation via `help` and I aim to utilise Kong instead but that's a stretch goal

sj26:
Interesting! What's stopping us from moving to Kong?

mcncl:
@sj26 they're completely different frameworks, we'd need to lift and shift to Kong entirely; we can't just move the `build` commands over, for example.

I'm really keen on doing it, and streamlining the output etc to be something useful like JSON by default rather than an interactive frame in the terminal.

sj26:
I asked Amp to have a go --

https://github.com/buildkite/cli/compare/kong

It's added kong as a parallel cli implementation when using `--use-kong`. 

I'm not sure if it's better? It's certainly rough, but an interesting proof of concept.

Anyway, that's probably not a blocker, yeah? We could still ship something like what @toolmantim proposed here? Are there any blockers to wrapping this with some tests and trying it out?

toolmantim:
I think just switching to Kong is the best thing to do, for humans and LLMs alike. I just opened a new PR for that.

PR #498, opened 2025-07-06
`bk configure` with org token already present

sj26:
I have an org token already configured:

```yaml
# ~/.config/bk.yaml
organizations:
    buildkite:
        api_token: ...
```

I go into a directory I haven't used bk in before, and it asks me to configure a token:

```
$ bk build view
Error: You must set a valid API token. Run `bk configure`. (failed during: bk)
```

So I do:

```
bk config
┃ Organization slug: buildkite
┃ API Token:
```

Hold on - it already knows my token! Why is it asking me, I don't know it!

We should skip this when we already have a token for the matching org. So, that's what this PR does.

(Technically the `bk build view` error is also incorrect - I have an API token configured, bk just doesn't understand which one to use.)

Based on https://github.com/buildkite/cli/pull/497 because Amp helped me out:

https://ampcode.com/threads/T-60c7ab55-fc0c-4c76-8855-3d8ba87fd8a0

### DIFF
--- a/AGENT.md
+++ b/AGENT.md
@@ -2,7 +2,7 @@ This project is the Buildkite CLI (`bk`)
 
 ## Commands
 - Test: `go test ./...`
-- Lint: `docker-compose -f .buildkite/docker-compose.yaml run lint`
+- Lint: `docker-compose -f .buildkite/docker-compose.yaml run golangci-lint golangci-lint run`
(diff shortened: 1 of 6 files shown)


mcncl (approved):

lizrabuya (approved):
🚀

PR #499, opened 2025-07-06
Handle missing TTY

### LINKED ISSUE #496: [Bug]: CLI fails if you try to use it when a TTY is not available
### Contact Details

_No response_

### What happened?

The CLI displays a spinner in some places, but this will fail with an error if a TTY is not present. We fixed this in one place in https://github.com/buildkite/cli/pull/409, but there are many other places where the standard spinner.New() is still being called. We should replace all those with SpinWhile, or perhaps consider discarding the spinner entirely. At any rate, there shouldn't be a hard requirement on a TTY.

### Version

3.8.0

### What environment are you seeing the problem on?

Local Development

### Relevant log output

```shell
Trying to run `bk build new` 

Aborted: user aborted: huh: could not open a new TTY: open /dev/tty: no such device or address (confirmation canceled (during: bk))
```

sj26:
I prompted Amp to remove the dependency on the TTY from the CLI. It's done a reasonable job!

It fixed a bug from my attempt to make spinners work out of TTY: #409

I'm not sold on its decision to always confirm in non-TTY. I think apt style "-y" which fails without confirmation by default might be better.

I also don't think it's done enough testing. But it's a great starting point!

Amp:

- Fix SpinWhile logic to properly handle non-TTY environments
- Add TTY detection to Confirm and PromptForOne functions
- Replace all spinner.New() calls with TTY-aware SpinWhile() calls
- Add non-TTY fallback for agent stop command using direct API calls
- Add comprehensive test coverage for TTY/non-TTY behavior
- All spinner operations now gracefully degrade when no TTY is available

Fixes #496

Amp-Thread: https://ampcode.com/threads/T-04774b89-5ae4-4839-830d-67c16e65acb6

### DIFF
--- a/internal/io/input.go
+++ b/internal/io/input.go
@@ -19,8 +19,11 @@ func HasDataAvailable(reader io.Reader) bool {
 	case *bufio.Reader:
 		return f.Size() > 0
 	case *strings.Reader:
-		return f.Size() > 0
+		// Check if there are unread bytes remaining
+		return f.Len() > 0
+	default:
+		// For other reader types, we can't easily determine if data is available
+		// This is a conservative approach - assume no data is available
+		return false
 	}
-
-	return false
 }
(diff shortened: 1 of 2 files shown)


mcncl:
@sj26 this looks good, did you want to rebase so we can merge in your PR?

sj26:
Extracted another little bit - just the clearer message when unconfirmed:

https://github.com/buildkite/cli/pull/503

PR #507, opened 2025-07-16
Switch CLI from Cobra to Kong framework

toolmantim:
Migrates from Cobra to Kong framework while adding structured output formats, comprehensive help with examples, and new job log management capabilities. This should help humans and LLMs better use the bk cli.

- Faster startup time (35ms → 8ms cold start)
- Structured output on all data commands (`-o json|yaml|table|raw`) for seamless scripting integration
- Machine-readable error output with stable exit codes and actionable recovery hints
- New `bk job logs` command for viewing individual job output and debugging failures
- Comprehensive help with examples on every command and full subcommand listing in top-level help
- One-command shell completions (`bk install-completions`) with kongplete for bash, zsh, fish, and PowerShell
- Flexible input parsing that accepts Buildkite URLs, slugs, or UUIDs across all commands
- Job log downloads via `--logs` flag with `--failed-only` filtering for faster debugging

All existing scripts continue to work with minimal changes. See UPGRADING.md for step-by-step upgrade instructions.

Replaces #478


mcncl:
@toolmantim that diff 😂 I'll look at this one file at a time, but I'm very excited if this does what it says on the tin

internal/http/client.go
```diff
@@ -163,6 +168,11 @@ func (c *Client) Do(ctx context.Context, method, endpoint string, body interface
 	}
 	req.Header.Set("Accept", "application/json")
 
+	// Set custom headers (these can override common headers)
```
toolmantim, line 171: custom headers support regressed in #447, this includes a fix for that

toolmantim:
@mcncl thanks! Lol yeah, no avoiding the big diff sadly. I just pushed a few final things, including a fix to custom headers. Feel free to update this branch yourself now.

pkg/cli/other.go
```diff
@@ -0,0 +1,855 @@
...
+		parts := strings.Split(strings.Trim(u.Path, "/"), "/")
+		if len(parts) >= 4 && parts[2] == "builds" {
+			return parts[0], parts[1], parts[3], nil
+		}
+		return "", "", "", fmt.Errorf("invalid build URL format")
+	}
+
+	// Build numbers without pipeline context require more complex resolution
+	// For now, require full URLs for simplicity
+	return "", "", "", fmt.Errorf("build number requires pipeline context - please use full build URL like: https://buildkite.com/org/pipeline/builds/%s", identifier)
```
scadu: We could previously use `bk build view $NUMBER`, now it expects full URI. Why's that?

pkg/cli/build.go
```diff
@@ -0,0 +1,825 @@
...
+	Build string `arg:"" help:"Build number (pipeline auto-detected from git repo) or build URL"`
+}
+
+type BuildDownloadCmd struct {
+	Build       string `arg:"" help:"Build number (pipeline auto-detected from git repo) or build URL"`
+	Destination string `help:"Download destination directory"`
+	Job         string `help:"Job ID to filter artifacts"`
+	Pattern     string `help:"Artifact name pattern"`
+	Logs        bool   `help:"Download job logs instead of artifacts"`
+	FailedOnly  bool   `help:"Only download logs from failed jobs (requires --logs)" name:"failed-only"`
```
scadu, line 43: Do we emit error/warning when `--failed-only` is passed without `--logs`, or fail silently?

pkg/cli/error_formatter.go
```diff
@@ -0,0 +1,163 @@
...
+	// Check if error is related to validation
+	msg := err.Error()
+	return contains(msg, "invalid") ||
+		contains(msg, "required") ||
+		contains(msg, "validation") ||
+		contains(msg, "format")
+}
+
+// contains checks if a string contains a substring (case-insensitive)
+func contains(s, substr string) bool {
```
scadu: Wouldn't something like `strings.Contains(strings.ToLower(s), substr)` be enough?

pkg/cli/headerflag.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/alecthomas/kong"
+)
+
+// HeaderFlag accumulates repeated --header values and supports both
+// colon format (KEY: VALUE) and equals format (KEY=VALUE) for compatibility
+type HeaderFlag struct {
+	Values map[string]string
+}
+
+func (h *HeaderFlag) Decode(ctx *kong.DecodeContext) error {
```
scadu, line 16: We might want to add tests to at least cover `Decode`.

pkg/cli/other.go
```diff
@@ -0,0 +1,855 @@
...
+	if conf.OrganizationSlug() == "" {
+		return fmt.Errorf("no organization selected. run `bk use` to select one")
+	}
+	return nil
+}
+
+// parseBuildIdentifier parses a build identifier which can be:
+// - A build URL (e.g., "https://buildkite.com/org/pipeline/builds/123")
+// - For now, build numbers without context are not supported
+func parseBuildIdentifier(identifier, defaultOrg string) (org, pipeline, buildNumber string, err error) {
```
scadu, line 27: Ditto about missing tests.

AGENT.md
```diff
@@ -12,7 +12,15 @@ This project is the Buildkite CLI (`bk`)
 ## Project Structure
 - Main binary: `cmd/bk/main.go`
 - GraphQL schema: `schema.graphql`
-- CLI commands: `pkg/cmd/`
+- CLI commands: `pkg/cmd/` (legacy Cobra implementation, unused)
```
scadu: If Cobra is gone, can we get rid of it? Also, with Cobra we had _some_ test coverage – not sure how much we have in place for Kong now? Still going through the PR 👀

pkg/cli/user.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				errChan <- fmt.Errorf("error creating user invite for %s: %w", email, err)
+			}
+		}(email)
+	}
+
+	go func() {
+		wg.Wait()
+		close(errChan)
+	}()
```
scadu: Nitpick: instead of spinning another goroutine with `go func` we can close the channel after `wg.Wait()`.

scadu (commented):
Still reading through the PR, but left some initial feedback.

pkg/cli/build.go
toolmantim: we error now, good catch!

AGENT.md
toolmantim: Gah, I swear I had fixed this… but no, looks like there's still some work to do that I missed. Will take a look.

toolmantim: Alright, just pushed up what should be the rest

toolmantim:
Related musings that I saw posted today: https://www.notcheckmark.com/2025/07/rethinking-cli-interfaces-for-ai/

I'll give this a really good manual test this week. If you can assist with code reviews, and someone can also give it a quick manual test, we should be good.

luispadron:
Was just about to open up an issue to allow `--json` output when possible (for use in scripts, etc). Would love to see this supported for commands like `bk build view` and `bk artifacts list`

main
scadu, line 1: Binary should be removed from the repo.

internal/config/config.go
```diff
@@ -111,8 +116,18 @@ func (conf *Config) SelectOrganization(org string, inGitRepo bool) error {
 func (conf *Config) APIToken() string {
 	slug := conf.OrganizationSlug()
 	key := fmt.Sprintf("organizations.%s.api_token", slug)
+
+	// Check for deprecated BK_ACCESS_TOKEN and warn if used
+	legacyToken := os.Getenv("BK_ACCESS_TOKEN")
```
scadu, line 121: Could it lead to a race condition when `sync.One` is a global, but warning logic depends on env variables that can change between calls?

internal/agent/list.go
```diff
@@ -56,36 +62,92 @@ func NewAgentList(loader func(int) tea.Cmd, page, perpage int) AgentListModel {
...
 func (m *AgentListModel) appendAgents() tea.Cmd {
+	return m.loadAgents(false)
+}
+
+func (m *AgentListModel) refreshAgents() tea.Cmd {
+	m.isRefreshing = true
+	return m.loadAgents(true)
+}
+
+func (m *AgentListModel) loadAgents(isRefresh bool) tea.Cmd {
```
scadu, line 73: Looks like it could append indefinitely in long-running sessions? Is there a limit or pagination somewhere?

pkg/cli/agent.go
```diff
@@ -0,0 +1,461 @@
...
+
+			org, agentID := parseAgentArg(agentArg, f.Config)
+
+			// Stop the agent
+			var err error
+			spinErr := bk_io.SpinWhile(fmt.Sprintf("Stopping agent %s", agentArg), func() {
+				_, err = f.RestAPIClient.Agents.Stop(ctx, org, agentID, a.Force)
+			})
+			if spinErr != nil {
+				errChan <- spinErr
```
scadu, line 247: Could it be another race condition when multiple goroutines write to `errChan` and reading from shared state? Might be worth looking into https://pkg.go.dev/golang.org/x/sync/errgroup which provides a friendly abstraction over goroutines handling.

internal/validation/scopes/scopes.go
scadu, line 1: Do we have equivalent of that in Kong version?

scadu (commented):
Left a few more comments, mostly around possible race conditions and validation.

PR #509, opened 2025-07-16
Remove main bin

toolmantim:
Similar to #508, it appears #498 introduced a binary called `main` that shouldn't be here


sj26:
Oh no my shame 🙈

PR #510, opened 2025-07-18
SUP-4182: Add --author flag to bk build new command

Mykematt:
- Add --author/-a flag to allow specifying build author
- Pass author parameter through newBuild function
- Set Author.Name field in CreateBuild struct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


Mykematt:
Committing a modification of the flag that accommodates all parameters for [author](https://pkg.go.dev/github.com/buildkite/go-buildkite/v4@v4.1.0#Author).

PR #511, opened 2025-07-19
SUP-4182: Add author flags to single --author flag with intelligent parsing to detect name, username, and email

Mykematt:
- Use parseAuthor() function to intelligently detect input type:
  - Email format (contains @ and .) → sets Email field
  - Full name (contains spaces) → sets Name field  
  - Username (default case) → sets Username field
- Supports both long (--author) and short (-a) forms
- Maintains full API compatibility with buildkite.Author struct

### DIFF
--- a/pkg/cmd/build/new.go
+++ b/pkg/cmd/build/new.go
@@ -160,7 +160,7 @@ func NewCmdBuildNew(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVarP(&message, "message", "m", "", "Description of the build. If left blank, the commit message will be used once the build starts.")
 	cmd.Flags().StringVarP(&commit, "commit", "c", "HEAD", "The commit to build.")
 	cmd.Flags().StringVarP(&branch, "branch", "b", "", "The branch to build. Defaults to the default branch of the pipeline.")
-	cmd.Flags().StringVarP(&author, "author", "a", "", "Author of the build (name, username, or email)")
+	cmd.Flags().StringVarP(&author, "author", "a", "", "Author of the build. Supports: \"Name <email>\", \"email@domain.com\", \"Full Name\", or \"username\"")
 	cmd.Flags().BoolVarP(&web, "web", "w", false, "Open the build in a web browser after it has been created.")
 	cmd.Flags().StringVarP(&pipeline, "pipeline", "p", "", "The pipeline to build. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}.\n"+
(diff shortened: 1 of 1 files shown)


mcncl (commented):
@Mykematt the REST API accepts an object for the `author` input, https://buildkite.com/docs/apis/rest-api/builds#create-a-build. How would the user go about providing both their name and their email? Does this only allow the setting of one of those things and what happens if they set `email` but not `name`, does `name` just autofill from an associated Buildkite account for the `email`?

Mykematt:
@mcncl Thanks for flagging this. The parseAuthor function has now been modified to accept both `name` and `email`. It also accepts either of them when given alone, and still accepts `username`. Below is the `--help` output from my local build of the modified code:
```
Create a new build on a pipeline.
The web URL to the build will be printed to stdout.

## To create a new build
$ bk build new

## To create a new build with environment variables set
$ bk build new -e "FOO=BAR" -e "BAR=BAZ"

## To create a new build with metadata
$ bk build new -M "key=value" -M "foo=bar"

Usage:
  bk build new [flags]

Flags:
  -m, --message string          Description of the build. If left blank, the commit message will be used once the build starts.
  -c, --commit string           The commit to build. (default "HEAD")
  -b, --branch string           The branch to build. Defaults to the default branch of the pipeline.
  -a, --author string           Author of the build. Supports: "Name <email>", "email@domain.com", "Full Name", or "username"
  -w, --web                     Open the build in a web browser after it has been created.
  -p, --pipeline string         The pipeline to build. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}.
                                If omitted, it will be resolved using the current directory.
  -e, --env stringArray         Set environment variables for the build
  -M, --metadata stringArray    Set metadata for the build (KEY=VALUE)
  -i, --ignore-branch-filters   Ignore branch filters for the pipeline
  -y, --yes                     Skip the confirmation prompt. Useful if being used in automation/CI
  -f, --env-file string         Set the environment variables for the build via an environment file
  -h, --help                    help for new

Global Flags:
  -V, --verbose   Enable verbose error output
```

mcncl:
@Mykematt thanks for the addition! Any tests we can add to this to ensure functionality, or do you have screenshots of;

- setting the various attributes
- setting an empty --author value
- ending the name with a space; "Baz Boo "
- that might do?

Mykematt:
Here are relevant screenshots:
Screenshot 2025-07-21 at 4 05 35 PM
Screenshot 2025-07-21 at 4 08 36 PM

PR #529, opened 2025-09-02
feat: implement `bk job list` command and enhance `bk agent list`

### LINKED ISSUE #527: 💡 feat: Add json output for bk agent list, to enable bulk pause/resume
### Is your feature request related to a problem?

Currently if we want to pause/resume agents, we can only do this one by one. And in production, multiple agents need to be paused at the same time.

### Describe the solution you'd like.

With json output, we can enable bulk action for agents.
For example for builds, we can list the builds based on filters first:
```
bk build list --pipeline test-pipeline --since 1h -o json
```
With `jq` and `xargs`, the build can be cancelled in bulk:
```
bk build list --pipeline test-pipeline --since 1h -o json \
| jq -r '.[].number' \
| xargs -r -n1 bk build cancel -p test-pipeline -y
```

### Describe alternatives you've considered.

_No response_

### Additional context

_No response_

### LINKED ISSUE #528: 💡 feat: List jobs based on queue name and other filters
### Is your feature request related to a problem?

The pipelines and queues are usually M:N mapping, and when alerts fire, engineers mostly knows what queue is having problem, and want to find out which job is causing the problem.
So it will be helpful to have `bk job list` functionality as well. If `-o json` output is supported, bulk cancel can also be enabled by jq and xargs.
It will be nice if we can have `--order-by` option, so the builds/jobs can be order by either start_time or duration.

### Describe the solution you'd like.

```
% bk job list --queue test-queue --state running --duration ">10m" --order-by duration
┌────────────────────────────────────────────────────────────────────────────────────────────────┬───────────────────────┬───────────────────────┬──────────┬──────────────┐
│ URL                                                                                            │ Started (UTC)         │ Finished (UTC)        │ State    │ Duration     │
├────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┼───────────────────────┼──────────┼──────────────┤
│ https://buildkite.com/company/ppl-pipeline/builds/1111111#019909ab-5f2e-41aa-9bd0-3797c32c7d2f │ 2025-09-02T09:04:12Z  │ -                     │ RUNNING  │ 11m 59s      │
│ https://buildkite.com/company/ppl-pipeline/builds/2222222#019909ab-5f45-41ac-98fe-a1f91bbc12ca │ 2025-09-02T09:04:13Z  │ -                     │ RUNNING  │ 11m 59s      │
│ https://buildkite.com/company/ppl-pipeline/builds/3333333#019909ab-5f32-44c0-a014-87698fff3355 │ 2025-09-02T09:04:14Z  │ -                     │ RUNNING  │ 11m 58s      │
└────────────────────────────────────────────────────────────────────────────────────────────────┴───────────────────────┴───────────────────────┴──────────┴──────────────┘
```

### Describe alternatives you've considered.

_No response_

### Additional context

_No response_

JoeColeman95:
Implements GitHub issue #528 with job listing functionality:
- Job filtering by queue, state, duration, pipeline, time range
- Hybrid server-side/client-side filtering for performance
- Sorting by start_time and duration with multiple output formats

Also enhances agent list with JSON/YAML output support and improved pagination.

Added test cases

Fixes #528
Fixes #527

### DIFF
--- a/pkg/cmd/job/list.go
+++ b/pkg/cmd/job/list.go
@@ -1,0 +1,523 @@
+package job
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/buildkite/cli/v3/internal/io"
+	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
+	"github.com/buildkite/cli/v3/internal/validation/scopes"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/cli/v3/pkg/output"
+	buildkite "github.com/buildkite/go-buildkite/v4"
+	"github.com/charmbracelet/lipgloss"
+	"github.com/spf13/cobra"
+)
+
+const (
+	maxJobLimit = 5000
+	pageSize    = 100
+)
+
+type jobListOptions struct {
+	pipeline string
+	since    string
+	until    string
+	duration string
+	state    []string
+	queue    string
+	orderBy  string
+	limit    int
+}
+
+func NewCmdJobList(f *factory.Factory) *cobra.Command {
+	var opts jobListOptions
+
+	cmd := cobra.Command{
+		DisableFlagsInUseLine: true,
+		Use:                   "list [flags]",
+		Short:                 "List jobs",
+		Long: heredoc.Doc(`
+			List jobs with optional filtering.
+
+			This command supports both server-side filtering (fast) and client-side filtering.
+			Server-side filters are applied when fetching builds, while client-side filters
+			are applied after extracting jobs from builds.
+
+			Client-side filters: --queue, --state, --duration
+			Server-side filters: --pipeline, --since, --until
+
+			Jobs can be filtered by queue, state, duration, and other attributes.
+			When filtering by duration, you can use operators like >, <, >=, and <= to specify your criteria.
+			Supported duration units are seconds (s), minutes (m), and hours (h).
+		`),
+		Example: heredoc.Doc(`
+			# List recent jobs (100 by default)
+			$ bk job list
+
+			# List jobs from a specific queue
+			$ bk job list --queue test-queue
+
+			# List running jobs
+			$ bk job list --state running
+
+			# List jobs that took longer than 10 minutes
+			$ bk job list --duration ">10m"
+
+			# List jobs from the last hour
+			$ bk job list --since 1h
+
+			# Combine filters
+			$ bk job list --queue test-queue --state running --duration ">10m"
+
+			# Order by duration (longest first)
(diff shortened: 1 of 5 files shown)


pkg/cmd/job/list.go
```diff
@@ -0,0 +1,523 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if len(builds) == 0 {
+			break
+		}
+
+		for _, build := range builds {
+			if len(allJobs)+len(build.Jobs) > cap(allJobs) {
```
scadu, line 192: I wonder if we can simplify it by using `append`?
Something like:
```
allJobs := make([]buildkite.Job, 0)
allJobs = append(allJobs, build.Jobs...)
```

pkg/cmd/job/list.go
```diff
@@ -0,0 +1,523 @@
...
+	cmd.Flags().StringVar(&opts.orderBy, "order-by", "", "Order results by field (start_time, duration)")
+	cmd.Flags().IntVar(&opts.limit, "limit", 100, fmt.Sprintf("Maximum number of jobs to return (max: %d)", maxJobLimit))
+
+	output.AddFlags(cmd.Flags())
+	cmd.Flags().SortFlags = false
+
+	return &cmd
+}
+
+func fetchJobs(ctx context.Context, f *factory.Factory, org string, opts jobListOptions, listOpts *buildkite.BuildsListOptions) ([]buildkite.Job, error) {
```
scadu, line 164: Could we simplify the job listing by using GraphQL? 
Cookbook for jobs: https://buildkite.com/docs/apis/graphql/cookbooks/jobs
This would also align with existing implementation where GraphQL is used, e.g. list queues, artifacts, .etc.

pkg/cmd/job/list.go
scadu: Although, we might want to consider using GraphQL here, if filtering jobs query wouldn't cut it: https://buildkite.com/docs/apis/graphql/cookbooks/builds

JoeColeman95: The 4 lines on code here could definitely be cleaned up with `append`, but I've consciously went with this method to avoid excessive memory allocations and slice copying (particularly when processing large datasets w/ thousands of jobs where performance will matter). This way is a bit more predictable 🔍

pkg/cmd/job/list.go
JoeColeman95: Did explore GraphQL on this one, but REST turned out to be a better fit. Jobs are union types (`JobTypeCommand`, `JobTypeWaiter`, etc.) which would mean dealing with fragments for each type, whereas REST gives us simple normalized structs.

The approach of filtering builds server-side (pipeline/since/until) then jobs client-side (queue/duration/state) works nicely with the REST endpoints, and simple page-based pagination fits our fetching pattern well. Gives us good performance on larger queries without overcomplicating things.

Issue #532, opened 2025-09-05
[Bug]: bk job list --queue option results in "No jobs found"


facingfrost:
### Contact Details

wanglinhan0530@gmail.com

### What happened?

```
% bk job list -p test-pipeline --state running
Jobs

State        Label                                  Started (UTC)        Finished (UTC)       Duration     URL
------------------------------------------------------------------------------------------------------------
running sleep                                  2025-09-05T12:26:28Z -                    26m41s (running) https://buildkite.com/company/pipeline/builds/number#id
% bk job list --queue test-queue --state running
No jobs found matching the specified criteria.
```
When running this command, there is a test job in progress, however, the filter doesn't show the correct build.
Also, it takes a long time to return the result(around 3m) with `--queue`. But with `-p` it's much faster.

### Version

3.11.0

### Relevant log output

```shell

```

Issue #531, opened 2025-09-05
💡 feat: bk job cancel command


facingfrost:
### Is your feature request related to a problem?

Similar to `bk build cancel`, `bk job cancel` will be very helpful.

### Describe the solution you'd like.

Similar to `bk build cancel`.

Issue #530, opened 2025-09-05
[Bug]: bk build list --creator option results in "No builds found"


facingfrost:
### Contact Details

wanglinhan0530@gmail.com

### What happened?

The creator filter for bk build list is returning the right build.
By running
```
% bk build list -p test-pipeline --state running
```
I get valid output:
```
Number   State        Message                   Started (UTC)        Finished (UTC)       Duration     URL
-------------------------------------------------------------------------------------------------------
9668     running test                      2025-09-05T12:22:29Z -                    8m59s (running) https://buildkite.com/company/test-pipeline/builds/number
```
However, when adding --creator filter, it's not returning builds, even if this build is triggered by myself.
```
% bk build list -p test-pipeline --state running --creator myemail@company.com
No builds found matching the specified criteria.
```
Could you please check this filter? Or I'm using the filter in a wrong way?

### Version

3.11.0

### Relevant log output

```shell
No builds found matching the specified criteria.
```

JoeColeman95:
Note: the text in this case is a bug, as this actually requires UUID (this is my fault as I opted for server-side filtering). So I will list this one as a bug and enhancement, as we will need to consider something like bk user list to get all users within an org, with additional filters to target specific attributes. That way we could look to do something like bk user list --email hello@world.com to return the object for that specific user. We could then filter with jq to get their UUID.

PR #538, opened 2025-09-11
Add email filtering, job queue filtering, and job cancel command

JoeColeman95:
Implements three GitHub issues with GraphQL optimizations:

- Email creator filtering (https://github.com/buildkite/cli/issues/530) `bk build list --creator alice@example.com` now resolves emails to user IDs via GraphQL

- Queue-based job filtering (https://github.com/buildkite/cli/issues/532) `bk job list --queue default` uses server-side GraphQL filtering for better performance

- Job cancel command (https://github.com/buildkite/cli/issues/531) `bk job cancel <job-id>` added following existing job command patterns

- Fixes `bk_io.Confirm()` function as it was not properly handling the `--yes` flag in non-interactive, even when provided it would still attempt to initialize a TTY causing failures. The benefits all confirmation prompts (build cancel, job cancel) and removes the requirement for TTY.

The job list implementation uses a hybrid approach of GraphQL for queue filtering, which prevents the performance issues with large datasets while maintaining full filter compatibility.

Fixes: https://github.com/buildkite/cli/issues/530
Fixes: https://github.com/buildkite/cli/issues/531
Fixes: https://github.com/buildkite/cli/issues/532

### DIFF
--- a/pkg/cmd/job/list.go
+++ b/pkg/cmd/job/list.go
@@ -324,14 +324,14 @@ func fetchQueuesFromClusters(ctx context.Context, client graphql.Client, cluster
 	resultChan := make(chan []string, len(clusters))
 	errorChan := make(chan error, len(clusters))
 	semaphore := make(chan struct{}, maxConcurrentRequests)
-	
+
 	var wg sync.WaitGroup
 
 	for _, cluster := range clusters {
 		wg.Add(1)
 		go func(c ClusterInfo) {
 			defer wg.Done()
-			
+
 			semaphore <- struct{}{}
 			defer func() { <-semaphore }()
 			queueIDs, err := fetchQueuesForCluster(ctx, client, c.ID, queueName)
@@ -339,7 +339,7 @@ func fetchQueuesFromClusters(ctx context.Context, client graphql.Client, cluster
 				errorChan <- fmt.Errorf("cluster %s: %w", c.Name, err)
 				return
 			}
-			
+
 			resultChan <- queueIDs
 		}(cluster)
 	}
@@ -347,21 +347,21 @@ func fetchQueuesFromClusters(ctx context.Context, client graphql.Client, cluster
 	var allQueueIDs []string
 	var results int
 	expectedResults := len(clusters)
-	
+
 	for results < expectedResults {
 		select {
 		case queueIDs := <-resultChan:
 			allQueueIDs = append(allQueueIDs, queueIDs...)
 			results++
-		
+
 		case err := <-errorChan:
 			return nil, err
-		
+
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		}
 	}
-	
+
 	return allQueueIDs, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/job/jobs.graphql
```diff
@@ -0,0 +1,99 @@
+query FindQueues($org: ID!) {
+  organization(slug: $org) {
+    clusters(first: 50) {
```
scadu: Looks like no pagination is used (the cursor is missing), and with hardcoded limit of 50, it might not cover larger orgs?

pkg/cmd/build/list.go
```diff
@@ -168,6 +184,40 @@ func NewCmdBuildList(f *factory.Factory) *cobra.Command {
...
+}
+
+func resolveCreatorEmailToUserID(ctx context.Context, f *factory.Factory, email string) (string, error) {
+	org := f.Config.OrganizationSlug()
+	resp, err := graphql.FindUserByEmail(ctx, f.GraphQLClient, org, email)
+	if err != nil {
+		return "", fmt.Errorf("failed to query user by email: %w", err)
+	}
+
+	if resp.Organization == nil || resp.Organization.Members == nil || len(resp.Organization.Members.Edges) == 0 {
```
scadu, line 199: Nitpick, but might be worth adding organization name. Helpful when user has access to multiple organizations.
Maybe something like:

```go
if resp.Organization == nil {
    return "", fmt.Errorf("organization not found: %s", org)
}
if resp.Organization.Members == nil || len(resp.Organization.Members.Edges) == 0 {
    return "", fmt.Errorf("no user found with email %s in organization %s", email, org)
}
```

JoeColeman95: I like the concern, but I'm not sure this adds much value since the CLI depends on `BUILDKITE_ORGANIZATION_SLUG` (config/env var) and API keys are scoped per org - so users already know which org context they're in when this error occurs based off of the config or any env vars passed with the command. If you still want it, I can add it. LMK what you think?

pkg/cmd/job/jobs.graphql
JoeColeman95: Good point - While 50 covers most typical cases, we should handle this properly. I've just pushed changes that implement pagination. We now fetch all clusters, then check queues against each cluster with cursor-based pagination 🙂

pkg/cmd/build/list.go
scadu: > [...] since the CLI depends on BUILDKITE_ORGANIZATION_SLUG

Forgot about that! True, it doesn't provide much values in this case.

pkg/cmd/job/list.go
```diff
@@ -112,13 +121,17 @@ func NewCmdJobList(f *factory.Factory) *cobra.Command {
 			var jobs []buildkite.Job
 
 			err = io.SpinWhile("Loading jobs", func() {
-				jobs, err = fetchJobs(cmd.Context(), f, org, opts, listOpts)
+				if opts.queue != "" {
+					jobs, err = fetchJobsWithQueueFilter(cmd.Context(), f, org, opts)
```
scadu, line 125: Would be good to document the queue filtering behavior when multiple clusters have queues with the same name.

PR #540, opened 2025-09-15
feat(build:list): add --no-limit option and fix duplicate builds for …

sbamaraa:
…non-divisible limits

### Add `--no-limit` Option:

Users can now specify `--no-limit` to fetch all builds that match the provided criteria.
The command will continue fetching until all available builds are retrieved, making it easier to work with large datasets or when the total number of builds is unknown.

### Fix for Non-Divisible Limits:

Previously, if the requested limit was not a multiple of the page size (e.g., 230 with a page size of 100), the command could return duplicate builds due to overlapping page boundaries.
The fetching logic has been corrected to ensure that only unique builds are returned, and the total count matches the requested limit.

### Acknowledge logic from user when fetching more than certain number of requests.
When user fetches maxBuildLimit number of builds through the API, now it asks acknowledgement.

### Loading spanner has more info
Now it shows how many raw builds have been fetched + how many actually matches the client side filter.

### Builds are displayed as soon as it is fetched
Instead of displaying all the builds all at once after the command is finished, now it prints them immediately.

### DIFF
--- a/pkg/cmd/build/list.go
+++ b/pkg/cmd/build/list.go
@@ -317,13 +317,17 @@ func fetchBuilds(cmd *cobra.Command, f *factory.Factory, org string, opts buildL
 			rawSinceConfirm = 0
 		}
 
-		io.SpinWhile(spinnerMsg, func() {
+		spinErr := io.SpinWhile(spinnerMsg, func() {
 			if opts.pipeline != "" {
 				builds, err = getBuildsByPipeline(ctx, f, org, opts.pipeline, listOpts)
 			} else {
 				builds, _, err = f.RestAPIClient.Builds.ListByOrg(ctx, org, listOpts)
 			}
 		})
+
+		if spinErr != nil {
+			return nil, spinErr
+		}
 
 		if err != nil {
 			return nil, err


pkg/cmd/build/list.go
```diff
@@ -259,9 +265,8 @@ func buildListOptionsFromFlags(opts *buildListOptions) (*buildkite.BuildsListOpt
 func fetchBuilds(ctx context.Context, f *factory.Factory, org string, opts buildListOptions, listOpts *buildkite.BuildsListOptions) ([]buildkite.Build, error) {
 	var allBuilds []buildkite.Build
 
-	for page := 1; len(allBuilds) < opts.limit; page++ {
+	for page := 1; opts.noLimit || len(allBuilds) < opts.limit; page++ {
```
JoeColeman95: This seems like it'd infinitely loop if there's any issue with the API, such as it returning the same results.

pkg/cmd/build/list.go
```diff
@@ -259,9 +265,8 @@ func buildListOptionsFromFlags(opts *buildListOptions) (*buildkite.BuildsListOpt
...
 		listOpts.Page = page
-		listOpts.PerPage = min(pageSize, opts.limit-len(allBuilds))
```
JoeColeman95, line 264: This will result in 30 results per page across the board which will reduce the efficiency quite dramatically

pkg/cmd/build/list.go
```diff
@@ -280,7 +285,19 @@ func fetchBuilds(ctx context.Context, f *factory.Factory, org string, opts build
...
+				break
+			}
+			if len(builds) > remaining {
+				allBuilds = append(allBuilds, builds[:remaining]...)
+			} else {
+				allBuilds = append(allBuilds, builds...)
+			}
+		} else {
+			allBuilds = append(allBuilds, builds...)
+		}
```
JoeColeman95: Potentially, this could be less heavy if we restored the page size, with some checks to ensure results are not duplicated?

pkg/cmd/build/list.go
```diff
@@ -177,6 +182,7 @@ func NewCmdBuildList(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVar(&opts.commit, "commit", "", "Filter by commit SHA")
 	cmd.Flags().StringVar(&opts.message, "message", "", "Filter by message content")
 	cmd.Flags().IntVar(&opts.limit, "limit", 50, fmt.Sprintf("Maximum number of builds to return (max: %d)", maxBuildLimit))
+	cmd.Flags().BoolVar(&opts.noLimit, "no-limit", false, "Fetch all builds (overrides --limit, sets perPage=500)")
```
JoeColeman95: Flagging this as the description does not match the implementation

eaugene:
@sbamaraa  I don't know how difficult it is to implement 
instead of --no-limit option , can we process to some records and then ask ack from user to get more data and process . ?

pkg/cmd/build/list.go
sbamaraa: Ok will resolve this problem.

pkg/cmd/build/list.go
sbamaraa: Ok will make the description match the implementation.

pkg/cmd/build/list.go
sbamaraa: When we change the number of items per page between requests, the pagination logic returns incorrect, duplicate data. For example, after successfully fetching the first 50 items on page 1, requesting the next 10 items on page 2 incorrectly gives us items 11 through 20 from the first page, instead of the expected items 51 through 60.

pkg/cmd/build/list.go
sbamaraa: Added the description here https://github.com/buildkite/cli/pull/540#discussion_r2351703265

sbamaraa:
> @sbamaraa I don't know how difficult it is to implement instead of --no-limit option , can we process to some MaxBuild records and then ask ack from user to get more data and process ? @JoeColeman95 thoughts about this approach to avoid infinite looping fetching all builds.

Would be happy to implement that logic, again any thoughts on it @JoeColeman95?

pkg/cmd/build/list.go
```diff
@@ -115,8 +116,12 @@ func NewCmdBuildList(f *factory.Factory) *cobra.Command {
 			// Get pipeline from persistent flag
 			opts.pipeline, _ = cmd.Flags().GetString("pipeline")
 
-			if opts.limit > maxBuildLimit {
-				return fmt.Errorf("limit cannot exceed %d builds (requested: %d)", maxBuildLimit, opts.limit)
+			if opts.noLimit {
+				opts.limit = 0
+			} else {
```
JoeColeman95:
```suggestion
			if !opts.noLimit {
 ```

File: pkg/cmd/build/list.go (hunk 4 above)
Comment (JoeColeman95) [inline pkg/cmd/build/list.go]:
```suggestion
	cmd.Flags().BoolVar(&opts.noLimit, "no-limit", false, "Fetch all builds (overrides --limit)")
```

pkg/cmd/build/list.go
```diff
@@ -259,9 +265,8 @@ func buildListOptionsFromFlags(opts *buildListOptions) (*buildkite.BuildsListOpt
 func fetchBuilds(ctx context.Context, f *factory.Factory, org string, opts buildListOptions, listOpts *buildkite.BuildsListOptions) ([]buildkite.Build, error) {
 	var allBuilds []buildkite.Build
 
-	for page := 1; len(allBuilds) < opts.limit; page++ {
+	for page := 1; opts.noLimit || len(allBuilds) < opts.limit; page++ {
 		listOpts.Page = page
```
JoeColeman95:
```suggestion
func fetchBuilds(ctx context.Context, f *factory.Factory, org string, opts buildListOptions, listOpts *buildkite.BuildsListOptions) ([]buildkite.Build, error) {
	var allBuilds []buildkite.Build
	var previousPageLastBuildNumber int

	listOpts.PerPage = pageSize

	for page := 1; ; page++ {
		if !opts.noLimit && len(allBuilds) >= opts.limit {
			break
		}

		listOpts.Page = page
```

pkg/cmd/build/list.go
JoeColeman95: Reimplemented in suggestion

pkg/cmd/build/list.go
JoeColeman95: Added tracker in suggestion to track to dupes, expands upon the code block further down to ensure there's no dupes

pkg/cmd/build/list.go
JoeColeman95: Adds page tracker

pkg/cmd/build/list.go
JoeColeman95: Added suggested change for this, then 🙂

pkg/cmd/build/list.go
JoeColeman95:
```suggestion
		if page > 1 && len(builds) > 0 {
			currentPageFirstBuildNumber := builds[0].Number

			if currentPageFirstBuildNumber == previousPageFirstBuildNumber {
				return nil, fmt.Errorf("API returned duplicate results, stopping to prevent infinite loop") // We should never get here
			}
		}

		if len(builds) > 0 {
			previousPageFirstBuildNumber = builds[0].Number
		}

		if !opts.noLimit {
			remaining := opts.limit - len(allBuilds)
			if remaining <= 0 {
				break
			}
			if len(builds) > remaining {
				allBuilds = append(allBuilds, builds[:remaining]...)
			} else {
				allBuilds = append(allBuilds, builds...)
			}
		} else {
			allBuilds = append(allBuilds, builds...)
		}
```

JoeColeman95: Keeps the original handling that you had here, but expands upon it to include the duplicate detection to prevent infinite loops

JoeColeman95:
@sbamaraa - I've added pagination back with my suggestions and reworked the logic slightly to add infinite loop prevention measures, and removed the 500 page size entry on the flag. The bulk of your changes were already great, so they remain untouched

sbamaraa:
I have made more changes to the PR :), additionally now it prints the results immediately after fetching. it helps debug much efficiently.

Also bunch of counter have been added to the "Loading spinner"

JoeColeman95:
Hey @sbamaraa, 

Looks great! 🙂 

The only thing we need here is the PageSize setting to 100 using the constant we have, as this will still use the default page size of 30 builds. We're getting a lint error due to the unchecked usage of Spinner:
```
pkg/cmd/build/list.go:320:15: Error return value of `io.SpinWhile` is not checked (errcheck)
```

JoeColeman95 (approved):

JoeColeman95:
Thanks for the contribution @sbamaraa 😄

PR #549, opened 2025-09-25
Migrate CLI entrypoint to main.go and Kong

JoeColeman95:
Moved the main CLI entry point from `cmd/bk/main.go` to `./main.go` and switched to Kong for command parsing.

## What's changed?
- CLI entry point is now at the repository root
- Using Kong parser with delegation to existing Cobra commands
- All existing functionality works exactly the same
- Moved `generate.go` to `cmd/generate/` to avoid conflicts

## Migration path
Later we can implement native Kong commands in separate files and remove the delegation layer. But for now this gets us the Kong foundation without changing any behavior. I've added comments to the sections we can rip out once we're ready to move on. I've created this with the assumption that we will continue to create the subcommands within their respective files.

Tested that all commands and help work the same as before.


main.go
```diff
@@ -0,0 +1,176 @@
...
+func (u *UseCmd) Run(*CLI) error       { return delegateToCobraSystem("use", u.Args) }
+
+func delegateToCobraSystem(command string, args []string) error {
+	originalArgs := os.Args
+	defer func() { os.Args = originalArgs }()
+
+	os.Args = append([]string{os.Args[0], command}, args...)
+
+	if code := runCobraSystem(); code != 0 {
+		os.Exit(code)
```
scadu, line 97: Do we have a way to show error message, too?

main.go
```diff
@@ -0,0 +1,176 @@
...
+		handleError(bkErrors.NewInternalError(err, "failed to create commands", "This is likely a bug", "Report to Buildkite"))
+		return bkErrors.ExitCodeInternalError
+	}
+
+	rootCmd.SetContext(context.Background())
+	rootCmd.SilenceErrors = true
+
+	if err := rootCmd.Execute(); err != nil {
+		handleError(err)
+		return 1
```
scadu, line 120: Should we use existing `bkErrors.GetExitCodeForError(err)`?
Maybe something like:
```go
func runCobraSystem() int {
    // existing code 
    if err := rootCmd.Execute(); err != nil {
        verbose, _ := rootCmd.PersistentFlags().GetBool("verbose")
        errHandler := bkErrors.NewHandler().WithVerbose(verbose)
        errHandler.HandleWithDetails(err, rootCmd.CommandPath())
        return bkErrors.GetExitCodeForError(err)
    }
    return 0
}
```

main.go
```diff
@@ -0,0 +1,176 @@
...
+	return 0
+}
+
+// We can rip this out when we have full Kong support for all commands
+func isHelpRequest() bool {
+	if len(os.Args) < 2 {
+		return false
+	}
+
+	// Global help, e.g. bk --help
```
scadu: We should also print help when no arguments are provided to match current behavior.

scadu (commented):
I like the hybrid approach, so we can keep existing functionality intact!
Left a few comments around error handling and help page.

JoeColeman95:
> This looks awesome, @JoeColeman95. I wonder should we use `kong.UsageOnError()` so that the help is outputted when only `bk` or `go run .` is used? Thoughts?

Hey @mcncl, yeah I agree – the functionality isn't quite as expected there though 🙁, see this issue: https://github.com/alecthomas/kong/issues/33

I've implemented the workaround stated there so we still get the expected behavior, I've delegated both `bk` and `bk --help` to Kong now and then anything `bk x [y z] --help` would result in Cobra help still.

Essentially, I've implemented exactly what you've suggested but added a condition to `run()` to catch and apply `--help` when there's no args passed on the base level

main.go
JoeColeman95: @scadu on this one the error messages are already being displayed, inside this function that you've shown a snippet of we're calling `handleError(err)` which uses the `bkErrors` package to format and display the message. The function then will return the relevant exit code. So by time we reach this part that you've highlighted, the error has already been shown to the user and we're just passing the exit code.

main.go
JoeColeman95: Hey @scadu on this one we have error handling here which is called during Kong-to-Cobra delegation before any Cobra command parsing happens so at this point we don't have access to the parsed flags like `verbose`.

If we were to implement your suggestion, it'd either fail or return default values since the command hasn't been executed yet. 

The current pattern is appropriate here as this is specifically for a delegation failures which is more for architectural errors than functional errors. The detailed error handling you are looking for here is still handled by the relevant sub-command within Cobra and remains untouched, so we'll still receive those errors directly from Cobra and have them passed back to Kong, for example:

```
./bk-test agent list --fdsdfsdf                                                                                                                                                                                        
Error: unknown flag: --fdsdfsdf
```

So in this setup we've got Kong handling top-level parsing and then Cobra handling command-specific errors with full context and the existing flags remaining present. This is so we can have the best of both worlds without adding new functionality that didn't exist prior, as we can do any of that once the migration is complete (such as --output being global, for example)

main.go
JoeColeman95: @scadu that's been added as part of the suggestion above from Ben but is handled from `run()` as this part will be scrapped once we have fully migrated

PR #552, opened 2025-10-01
⚡️ perf: optimize pipeline filtering by repository in API calls

abhishekmukherg:
Replace local filtering with repository-specific API requests to reduce data transfer and improve performance when filtering pipelines by repository URLs.

I have absolutely no idea if this is a good thing or not in the grand scheme of things, but as it stands now, `bk build view` is useless to me as it tries to enumerate every pipeline we own at our organization, which is... a lot. This leverages the repo url of the current repository to filter what's being listed, which makes it a LOT faster.

I can view this as being problematic if the use case involves forks? But I figured I'd throw this up there in case it sounds interesting to anyone.

### DIFF
--- a/internal/pipeline/resolver/repository.go
+++ b/internal/pipeline/resolver/repository.go
@@ -52,30 +52,33 @@ func filterPipelines(ctx context.Context, repoURLs []string, org string, client 
 	var currentPipelines []pipeline.Pipeline
 	page := 1
 	per_page := 30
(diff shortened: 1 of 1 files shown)


mcncl:
Thanks @abhishekmukherg! This kind of optimisation was on our roadmap so we'll take a look and see if at can merge it in

lizrabuya (approved):
Looks good to me!

PR #558, opened 2025-10-30
Add a migrate command for pipelines

mcncl:
## Description

With our work on implementing a better backend for the migration API, we can probably support a command in `pipeline` to allow folks to translate locally.

## Changes

- Adds a `migrate` subcommand to `pipeline`
- Adds integration testing for the command
- Allows overwrite of API endpoint, sets via LD flags
- Skips validation of an organisation slug or token (no need for a local migration)

### DIFF
--- a/pkg/cmd/pipeline/migrate_test.go
+++ b/pkg/cmd/pipeline/migrate_test.go
@@ -43,8 +43,8 @@ jobs:
 		t.Error("Expected job ID to be returned")
 	}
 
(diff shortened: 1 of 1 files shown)


PriyaSudip:
Nice addition @mcncl. Should we test the migrated pipeline in Buildkite by running it or leave it for customers?

PriyaSudip (approved):

PR #562, opened 2025-11-11
SUP-5203: use keychain for api token

mcncl:
## Description

Currently we store API tokens in plain text in a config file in the user's `.config` directory, that's not ideal.

## Changes

This change moves to using keychain (or Linux/Windows variants) for API tokens by default, but users can override this by setting the token source as `file` using an environment variable.

We've implemented the following;

- `bk configure add` still prompts as normal, but on storing the API token we use the `organization` as the identifier within keychain
- provides a `bk configure migrate` command to move existing tokens and their organization to keychain
- removes the use of a `~/.config/bk.yaml` file entirely
- tests included for this change (all passing)


JoeColeman95 (approved):
LGTM - Only thing is I'd put something about the keychain org list limitation in the heredoc somewhere

scadu:
Some folks use the CLI in CI pipelines, so we definitely want to mention that as a breaking chance, since keyring would be the default now.

mcncl:
@scadu it should be falling back to the config if the token is not present in keychain

internal/config/config.go
```diff
@@ -108,35 +119,70 @@ func (conf *Config) SelectOrganization(org string, inGitRepo bool) error {
 }
 
 // APIToken gets the API token configured for the currently selected organization
+// Priority order: BUILDKITE_API_TOKEN env var → keychain → config file (legacy)
 func (conf *Config) APIToken() string {
+	// Environment variable takes precedence
```
mcncl, line 124: @scadu this section should be what you're after

scadu: @mcncl ah, thanks for pointing that! Nice!

mcncl:
@JoeColeman95 good point; `bk use` won't work with this if a user does have more than one org, they'd need to run `bk configure add` and then provide their org which already exists in Keychain for it to then "use" that org

I'll work on a fix; storing orgs as keys with empty values in the config file so the list of user orgs is still present and can be used as a look up against Keychain

mcncl:
Going to close this while I work on some other things which will make this smoother

PR #563, opened 2025-11-18
SUP-4495, SUP-4957, SUP-4958, SUP-4956, SUP-4959, SUP-4961: Added global --yes and --no-input

Mykematt:
- Replace Huh prompts with simple `fmt.Scanln` for Y/N confirmations
- Replace Huh select menus with numbered list selection
- Add global `--yes` / `-y` flag to skip all confirmations
- Add global `--no-input` flag for strict automation mode
- Remove individual `--yes` flags from `job cancel`, `build new`, and `build cancel` 
- Add `Factory.SetGlobalFlags()` helper for command implementations
Screenshot 2025-11-17 at 5 14 03 PM
Screenshot 2025-11-17 at 5 15 37 PM

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ type CLI struct {
 	Use       UseCmd       `cmd:"" help:"Select an organization"`
 	User      UserCmd      `cmd:"" help:"Invite users to the organization"`
 	Version   VersionCmd   `cmd:"" help:"Print the version of the CLI being used"`
+	Whoami    WhoamiCmd    `cmd:"" help:"Print the current user and organization"`
 }
 
 // Hybrid delegation commands, we should delete from these when native Kong implementations ready
@@ -70,6 +71,9 @@ type (
 	UseCmd struct {
 		Args []string `arg:"" optional:"" passthrough:"all"`
 	}
+	WhoamiCmd struct {
+		Args []string `arg:"" optional:"" passthrough:"all"`
+	}
 )
 
 // Delegation methods, we should delete when native Kong implementations ready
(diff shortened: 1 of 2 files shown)


internal/io/confirm.go
```diff
@@ -1,30 +1,36 @@
...
+//	    // ... rest of your logic
+//	}
+//
+// Then use factory flags in RunE:
+//
+//	confirmed := f.SkipConfirm  // Respects global --yes flag
+//	io.Confirm(&confirmed, "Do the thing?")
+//
+// Do NOT add individual --yes flags to commands. Use the global flag pattern above.
 func Confirm(confirmed *bool, title string) error {
```
moskyb: i generally try to avoid modifying pointer arguments if at all possible — i wonder if we could change the signature up to be a little more ergonomic to go

```Go
func Confirm(factory *factory.Factory, prompt string) (bool, error) {
	if factory.SkipConfirm {
		return true, nil
	}

	fmt.Printf("%s [y/N]: ", prompt)
	var response string
	_, err := fmt.Scanln(&response)
	if err != nil {
		return false, fmt.Errorf("scanning for confirmation input: %w", err)
	}

	response = strings.ToLower(strings.TrimSpace(response))
	return response == "y" || response == "yes", nil
}
```

moskyb: in this case we could add the `NoInput` in there too actually, eg
```Go
// ...
if factory.NoInput {
  return false, errors.New("interactive input required but --no-input is set")
}
```

which means that callers have to do less thinking when they prompt the user

Mykematt: Thanks @moskyb. Fixed accordingly.

PR #569, opened 2025-11-24
SUP-4496, SUP-4951, SUP-4952, SUP-4954, SUP-4955: Remove Spinner Dependencies

Mykematt:
- Added global --quiet flag to suppress progress output 
- Stored in Factory.Quiet and propagated to commands 
- Removed charmbracelet/huh/spinner dependency 
- New simple spinner using ... animation, writes to stderr 
- Respects quiet mode and non-TTY environments 
- SpinWhile() now takes *factory.Factory parameter 
- Removed spinner from StoppableAgent bubbletea model 
- Handles non-TTY stdin properly for piped input 
- Job cancel: passes display ID (not GraphQL ID) to spinner message

### DIFF
--- a/cmd/build/cancel.go
+++ b/cmd/build/cancel.go
@@ -40,6 +40,7 @@ func (c *CancelCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 
 	f.SkipConfirm = globals.SkipConfirmation()
 	f.NoInput = globals.DisableInput()
+	f.Quiet = globals.IsQuiet()
 
 	if err := validation.ValidateConfiguration(f.Config, kongCtx.Command()); err != nil {
 		return err
--- a/cmd/build/create.go
+++ b/cmd/build/create.go
@@ -56,6 +56,7 @@ func (c *CreateCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 
 	f.SkipConfirm = globals.SkipConfirmation()
 	f.NoInput = globals.DisableInput()
+	f.Quiet = globals.IsQuiet()
 
 	if err := validation.ValidateConfiguration(f.Config, kongCtx.Command()); err != nil {
 		return err
--- a/cmd/build/download.go
+++ b/cmd/build/download.go
@@ -54,6 +54,7 @@ func (c *DownloadCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error 
 
 	f.SkipConfirm = globals.SkipConfirmation()
 	f.NoInput = globals.DisableInput()
+	f.Quiet = globals.IsQuiet()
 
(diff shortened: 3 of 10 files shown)


pkg/cmd/build/cancel.go
```diff
@@ -28,7 +28,20 @@ func NewCmdBuildCancel(f *factory.Factory) *cobra.Command {
 			Cancel the given build.
 		`),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			f.SetGlobalFlags(cmd)
+			// Get the command's required and optional scopes
+			cmdScopes := scopes.GetCommandScopes(cmd)
+
+			// Get the token scopes from the factory
+			tokenScopes := f.Config.GetTokenScopes()
```
scadu: Not used anymore, see https://github.com/buildkite/cli/pull/566

Mykematt: Thanks @scadu. Fixed now!

internal/io/spinner.go
```diff
@@ -1,22 +1,44 @@
...
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 	"github.com/mattn/go-isatty"
 )
 
-func SpinWhile(name string, action func()) error {
-	if !isatty.IsTerminal(os.Stdout.Fd()) {
-		// No TTY available, just run the action without spinner
+func SpinWhile(f *factory.Factory, name string, action func()) error {
+	// If quiet mode is on or not a terminal, just run the action
+	if f.Quiet || !isatty.IsTerminal(os.Stderr.Fd()) {
```
scadu: Should we keep using `stdout`? It looks like other places in the codebase rely on `stdout`.

internal/agent/list.go
```diff
@@ -58,13 +58,12 @@ func NewAgentList(loader func(int) tea.Cmd, page, perpage int) AgentListModel {
 func (m *AgentListModel) appendAgents() tea.Cmd {
 	// Set agentsLoading
 	m.agentsLoading = true
-	// Set a status message and start the agentList's spinner
-	startSpiner := m.agentList.StartSpinner()
+	// Set a status message
 	statusMessage := fmt.Sprintf("Loading more agents: page %d of %d", m.agentCurrentPage, m.agentLastPage)
 	setStatus := m.agentList.NewStatusMessage(statusMessage)
```
scadu, line 73: Looks like `quiet` is not propagated here (and probably other places aside from `SpinWhile`. Might be a good idea to add a flag to the `api` layer to suppress the output.

internal/io/spinner.go
Mykematt: Fixed

internal/agent/list.go
Mykematt: Applied too. `--quiet` has been moved to the `api` layer and removed from specific command level.

PR #570, opened 2025-11-25
Prompt user to provide a cluster when creating a pipeline

lizrabuya:
The create pipeline [API](https://buildkite.com/docs/apis/rest-api/pipelines#create-a-yaml-pipeline) now requires a cluster_id to create the pipeline. The `bk cli` will now prompt the user to select from the org's existing clusters, or skip when the org was created with only an [unclustered area](https://buildkite.com/docs/pipelines/clusters/migrate-from-unclustered-to-clustered-agents#buildkite-organizations-created-after-february-26-2024).

### DIFF
--- a/pkg/cmd/pipeline/create.go
+++ b/pkg/cmd/pipeline/create.go
@@ -3,6 +3,7 @@ package pipeline
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	survey "github.com/AlecAivazis/survey/v2"
 	"github.com/MakeNowJust/heredoc"
@@ -44,29 +45,33 @@ func NewCmdPipelineCreate(f *factory.Factory) *cobra.Command {
 				return err
 			}
 
-			clusterMap := getClusters(cmd.Context(), f)
-			clusterNames := make([]string, 0, len(clusterMap))
+			clusterMap, err := getClusters(cmd.Context(), f)
+			if err != nil {
+				return err
+			}
 
+			clusterNames := make([]string, 0, len(clusterMap))
 			for name := range clusterMap {
 				clusterNames = append(clusterNames, name)
 			}
 			sort.Strings(clusterNames)
-			if len(clusterNames) > 0 {
+			clusterNames = append([]string{"Skip (No Cluster)"}, clusterNames...)
+
+			if len(clusterMap) > 0 {
+
 				prompt := &survey.Select{
 					Message: "Choose a cluster:",
 					Options: clusterNames,
 				}
+
 				var selectedClusterName string
 				err := survey.AskOne(prompt, &selectedClusterName, survey.WithValidator(survey.Required))
 				if err != nil {
 					return err
 				}
-				clusterID = clusterMap[selectedClusterName]
-			} else {
-				// Use user provided answer as the cluster ID. This is currently optional since we have orgs that still allow pipelines without clusters
-				err := survey.AskOne(&survey.Input{Message: "No clusters found. Optionally provide a Cluster ID (press Enter to skip):"}, &clusterID)
-				if err != nil {
-					return err
+
+				if selectedClusterName != "Skip (No Cluster)" {
+					clusterID = clusterMap[selectedClusterName] // will be "" if "Skip (no cluster)" was selected
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/pipeline/create.go
```diff
@@ -43,6 +44,31 @@ func NewCmdPipelineCreate(f *factory.Factory) *cobra.Command {
...
 			}
 
+			clusterMap := getClusters(cmd.Context(), f)
+			clusterNames := make([]string, 0, len(clusterMap))
+
+			for name := range clusterMap {
+				clusterNames = append(clusterNames, name)
+			}
+			if len(clusterNames) > 0 {
+				prompt := &survey.Select{
```
scadu, line 62: `survey` is part of charmbracelet/bubbletea which we plan to stop using as part of the CLI modernization initiative, so we might want to find an alternative. t

JoeColeman95: On this point, I think that we will need to review this overall as we need to explore automation-first workflows, can see this subcommand and `bk cluster list`, `bk configure add`, etc will need to be overhauled. Not an issue as the premise of this PR is just adding additional functionality IMO.

pkg/cmd/pipeline/create.go
JoeColeman95, line 6: Add "sort" here so we can do sort.Strings below

pkg/cmd/pipeline/create.go
```diff
@@ -43,6 +44,31 @@ func NewCmdPipelineCreate(f *factory.Factory) *cobra.Command {
 				return err
 			}
 
+			clusterMap := getClusters(cmd.Context(), f)
+			clusterNames := make([]string, 0, len(clusterMap))
+
+			for name := range clusterMap {
+				clusterNames = append(clusterNames, name)
+			}
```
JoeColeman95, line 56:
```suggestion
			for name := range clusterMap {
				clusterNames = append(clusterNames, name)
			}
			sort.Strings(clusterNames)
```

JoeColeman95: Minor/UX: This here will ensure that the list is always in the same order. This only really applies to orgs that have a lot of clusters, but nice UX. You'll need to import sort at the top of the page for this to work 🙂

This is due to the way that maps work in Go being random: https://go.dev/blog/maps#iteration-order

JoeColeman95: Example of this can be seen here:

Image

pkg/cmd/pipeline/create.go
```diff
@@ -83,7 +109,24 @@ func getRepoURLS(f *factory.Factory) []string {
...
+	if len(clusters) < 1 {
+		return map[string]string{}
+	}
+
+	clusterMap := make(map[string]string, len(clusters))
+	for _, cluster := range clusters {
+		clusterMap[cluster.Name] = cluster.ID
+	}
+	return clusterMap
+}
```
JoeColeman95: Blocking: We should paginate here as this will only fetch the first page of results which defaults to 30. In this case, it'd be widely working, but some users may have more than 30 clusters, which would mean they'd only see some of the results.

We have pagination in a few places, but it can vary quite a lot depending on the call. The best example of this I can see in this code base is in the [filterPipelines](https://github.com/buildkite/cli/blob/fix/add_cluster_on_create_pipeline/internal/pipeline/resolver/repository.go#L51-L85) function, which should be 1:1 transferable here.

pkg/cmd/pipeline/create.go
```diff
@@ -83,7 +109,24 @@ func getRepoURLS(f *factory.Factory) []string {
 	return c.Remotes["origin"].URLs
 }
 
-func createPipeline(ctx context.Context, client *buildkite.Client, org, pipelineName, description, repoURL string) error {
+func getClusters(ctx context.Context, f *factory.Factory) map[string]string {
+	clusters, _, err := f.RestAPIClient.Clusters.List(ctx, f.Config.OrganizationSlug(), nil)
+	if err != nil {
+		return map[string]string{}
```
JoeColeman95: Minor: API errors are silently swallowed here. Not blocking, but could be confusing if the list call fails for network/auth reasons

pkg/cmd/pipeline/create.go
```diff
@@ -43,6 +44,31 @@ func NewCmdPipelineCreate(f *factory.Factory) *cobra.Command {
...
+			for name := range clusterMap {
+				clusterNames = append(clusterNames, name)
+			}
+			if len(clusterNames) > 0 {
+				prompt := &survey.Select{
+					Message: "Choose a cluster:",
+					Options: clusterNames,
+				}
+				var selectedClusterName string
+				err := survey.AskOne(prompt, &selectedClusterName, survey.WithValidator(survey.Required))
```
JoeColeman95, line 68: Design choice/blocking: This seems like it could be breaking for some users. The API at the moment sets `cluster_id` to optional, in this case, we're forcing an option if there is a cluster available, but what if an org has a cluster with 0 agents but 1+ unclustered agent(s)? We'd be forcing them to select an unusable cluster, should we allow skipping even if there's a cluster in this case to allow these cases? I think respecting the API is the best bet.

pkg/cmd/pipeline/create.go
lizrabuya: Yeah, same thoughts with @JoeColeman95  here. The PR was just to add the functionality as this will block users with newly created orgs. Removing any part of `charmbracelet/bubbletea` should be covered differently.

pkg/cmd/pipeline/create.go
lizrabuya: Good catch! Yeah, I agree we should allow skipping even if there is a cluster available.

JoeColeman95 (approved):
Very nice! 😄

PR #572, opened 2025-12-02
fix: migrate api url

mcncl:
## Description

The migrate API has it's own subdomain now, so we should use it.

## Changes

- Points to the `migrate.` instead of API gateway URL


mcncl:
Hm, there's a chance it needs "production" appended. I'll look tomorrow

PR #575, opened 2025-12-05
SUP-4172: Preserve query parameters in bk api requests

Mykematt:
Previously, query parameters like `/builds?branch=main` were being URL-encoded (?→%3F), causing 404 errors. The fix separates the query string before `url.JoinPath()` and reattaches it after.

Fixed at the HTTP level instead of `api` command level (`api.go`), so no need for flags to pass through each url (sub)path:
- `internal/http/client.go` - Parse and preserve query parameters in URL construction
- `internal/http/client_test.go` - Added tests for query parameter handling

Fix at the `api` command level would have been: `bk api /builds --field branch=main --field state=passed`
Implemented fix at the HTTP level is: `bk api "/builds?branch=main&state=passed"`

This approach also saves refactoring code to match kong when the `api` command eventually gets its own migration to kong from cobra


mcncl (commented):
I think this would work if the URL is in an ideal state, but would fail if incorrect/unexpected encoding was used and/or missing?

For example, using the `api` call with `/builds?branch=feature/test name&state=passed`, I think would fail due to the space?

I think that would also be the case if any fragments were passed in using `#`? Ie, what would happen if the user sets the API call to `/builds?branch=main#foo`?

There is a `Parse` function available on the `url` package of `stdlib` which I think would do a much better job than using the `strings` package.

mcncl (approved):

PR #582, opened 2025-12-09
Migrating Artifacts Commands to Kong

JoeColeman95:
Migrates bk artifacts download and list from Cobra to Kong and adds exception in delegation for help to cobra.

Moved artifacts graphQL schema to new cmd/artifacts/graphql folder

Standardized the import of bkIO across all files for consistency, all works fine still

Tested locally, all works as previously

### DIFF
--- a/cmd/build/view.go
+++ b/cmd/build/view.go
@@ -141,7 +141,7 @@ func (c *ViewCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 	var wg sync.WaitGroup
 	var mu sync.Mutex
 
-		spinErr := bkIO.SpinWhile(f, "Loading build information", func() {
+	spinErr := bkIO.SpinWhile(f, "Loading build information", func() {
 		wg.Add(3)
 		go func() {
 			defer wg.Done()


cmd/artifacts/download.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/alecthomas/kong"
+	"github.com/buildkite/cli/v3/internal/cli"
+	bkGraphQL "github.com/buildkite/cli/v3/internal/graphql"
+	bkIO "github.com/buildkite/cli/v3/internal/io"
```
scadu, line 14: Curious/nit: we already import `internal/io` under a few other prefixes: `bk_io`, `bkio` and `bkIO` is another one.

JoeColeman95: @scadu yeah, you're right – in this case, I went with bkIO but for a lot of the others it was mainly a case of them being pre-existing and not standardised in naming. In this case bkIO seemed rational to me with bkGraphQL above it. I'll push a fix for this to standardise everything to bkIO as it's purely cosmetic

PR #583, opened 2025-12-09
Migrating cluster commands to Kong

JoeColeman95:
Migrated the cluster commands from Cobra to Kong

Added signal aware context to allow the command to be gracefully cancelled as this would of been background otherwise (could be messy for a lot of clusters), only change worth noting as would of been a little worse moving to Kong otherwise due to change in context.

Tested as working locally

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,8 +105,8 @@ type (
 )
 
 // Delegation methods, we should delete when native Kong implementations ready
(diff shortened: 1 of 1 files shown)


JoeColeman95:
#582 I accidentally pushed list.go, this is non-breaking, but worth noting as it's not showing here on this PR. This is why we do 1 at a time 🥇

PR #585, opened 2025-12-10
[cli/api] Add automatic retry with backoff for 429 rate limit responses

thomasdesr:
## Summary
- This PR adds (& enables by default) automatic retry for HTTP `429 (Too Many Requests)` the `bk api` commands
- Added `--verbose` flag to observe retry behavior when using the CLI

Note: I didn't want to make this the default anywhere other than the CLI because I don't know what the consequences would be of making this the default 🙃 If you think its safe, happy to make it the default everywhere 🤣 

## Implementation Details
- Built following the docs here: https://buildkite.com/docs/apis/rest-api/limits
- When the client experiences a `429` it won't retry before `RateLimit-Reset` timer has expired
- If after that, it _still_ hits another `429`, it starts to exponentially backoff on repeated rate limits
- `internal/http` gained a few additional options to configure the behavior:
  - `WithMaxRetries` to control how many attempts & `WithMaxRetryDelay()` to set an upper bound on retry intervals 
  - `WithOnRetry` callback (currently used by the `--verbose` flag for printing warnings about backoff being hit)

### DIFF
--- a/cmd/agent/list.go
+++ b/cmd/agent/list.go
@@ -71,7 +71,7 @@ Examples:
 }
 
 func (c *ListCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
-	f, err := factory.New()
+	f, err := factory.New(factory.WithDebug(globals.EnableDebug()))
 	if err != nil {
 		return err
 	}
--- a/cmd/agent/pause.go
+++ b/cmd/agent/pause.go
@@ -40,7 +40,7 @@ Examples:
 }
 
 func (c *PauseCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
-	f, err := factory.New()
+	f, err := factory.New(factory.WithDebug(globals.EnableDebug()))
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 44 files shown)


thomasdesr:
Hey @scadu! Thanks for the second look! I've rebased onto main and resolved conflicts from the Kong migration.

Two thoughts:
* I re-added the `--verbose` flag into the ApiCmd Kong struct, but kept it command-local for now rather than root-level, lmk if you want me to move it back to global! I saw it wasn't setup yet at the global level
* I'm also not super familiar with roko so if you could take a close look at how I'm using it, I'd appreciate it :D Would love confirmation if `roko.Constant(0) + SetNextInterval()` is the right pattern for server-directed delays, or if there's a more idiomatic approach 🙏 

Sorry for the turnaround delay! Ptal :D

thomasdesr:
> In this case we can consider using roko.Exponential, e.g.: roko.WithStrategy(roko.Exponential(2, 0)), // Wait (2 ^ attemptCount) + 0 seconds between attempts.

I don't think this actually works due to how `SetNextInterval` works, it'll override the next attempt's sleep right? And since we call it every time, it will never actually hit the strategy's "sleep amount".

Please let me know if I'm misunderstanding either its behavior or your suggest 🙏

PR #588, opened 2025-12-10
feat(job:list): add --no-limit option

scadu:
### Description

When using client-side filters (`--duration`, `--state`, `--queue`), the default behavior only fetches a limited number of builds, which can cause matching jobs to be missed. 
This adds a `--no-limit` flag (consistent with build list added in #540) to allow users to explicitly fetch all builds when 

Default: fetches up to `limit*2` 
`--no-limit`: fetches all builds

Example:
`bk job list --duration 1m --limit 10`
      -> 4 results (from ~20 builds)

`bk job list --duration 1m --limit 10 --no-limit`
      -> 4179 results (from all builds)

### DIFF
--- a/cmd/job/list.go
+++ b/cmd/job/list.go
@@ -49,9 +49,8 @@ are applied after extracting jobs from builds.
 Client-side filters: --queue, --state, --duration
 Server-side filters: --pipeline, --since, --until
 
-IMPORTANT: When using client-side filters (--duration, --state, --queue), you may need
-to use --no-limit to ensure all matching jobs are found, as the default behavior only
-fetches a limited number of builds.
+By default, fetches up to 200 builds for filtering. Use --no-limit if you need to
+search across more builds to find all matching jobs.
 
 Jobs can be filtered by queue, state, duration, and other attributes.
 When filtering by duration, you can use operators like >, <, >=, and <= to specify your criteria.
@@ -190,8 +189,9 @@ func fetchJobs(ctx context.Context, f *factory.Factory, org string, opts jobList
 		// When --no-limit is set, fetch all available builds (no upper bound)
 		maxBuildsToFetch = 0 // 0 means unlimited
 	} else {
-		// By default, be conservative and only fetch what we need based on limit
-		maxBuildsToFetch = min(200, opts.limit*2)
+		// By default, fetch a reasonable number of builds (200 = 2 pages)
+		// This provides a good pool for filtering without being tied to --limit
+		maxBuildsToFetch = 200
 	}
 
 	allJobs := make([]buildkite.Job, 0, opts.limit*2)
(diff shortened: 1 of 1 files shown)


cmd/job/list.go
```diff
@@ -174,14 +175,28 @@ func (c *ListCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
...
+	// we have enough jobs after filtering. Otherwise, we can optimize by only
+	// fetching what we need based on the limit.
+	hasClientSideFilters := len(opts.state) > 0 || opts.duration != ""
+
+	var maxBuildsToFetch int
+	if hasClientSideFilters {
+		// When filtering client-side, fetch all available builds up to a reasonable maximum
+		// to ensure we don't miss matching jobs. Use maxJobLimit (5000) as the upper bound.
+		maxBuildsToFetch = maxJobLimit
+	} else {
```
JoeColeman95: Fetching up to maxJobLimit (5k jobs) for any client-side filter feels like a bad default as it could hammer the API and affect latency at scale (specifically when tools such as parallel are used). Could we cap more conservatively here by scaling with limit like `min(maxJobLimit, max(200, limit*10))` instead? Or potentially introduce a configurable fetch cap? A `--no-limit` would make the "fetch everything" behaviour explicit rather than implicit?

cmd/job/list.go
```diff
@@ -543,6 +567,18 @@ func jobListOptionsFromFlags(opts *jobListOptions) (*buildkite.BuildsListOptions
...
+	// Note: This filters builds by state, but we'll still apply exact job-level
+	// state filtering client-side since build state != job state
+	if len(opts.state) > 0 {
+		// Map job states to build states where they align
+		// Job states like "passed", "failed", "running", etc. map reasonably to build states
+		listOpts.State = make([]string, len(opts.state))
+		for i, state := range opts.state {
+			listOpts.State[i] = strings.ToLower(state)
+		}
+	}
```
JoeColeman95: This seems unreasonable, and would be a regression in the tooling as job states ≠ build states and in this context could result in builds being dropped when they contain jobs that we want as results. I'd suggest dropping this concept entirely as there's no real win unless we see improvements to the jobs api directly.

JoeColeman95 (changes requested):
Left comments on 2 major issues/concerns here, I get the concept behind the PR (increase the sample) but we're inherently limited by the API itself in this case, and note the filtering types to make this more apparent.

I suspect we could just detach the `--limit` flag from the number of results we query (although that was the intended design), or even simply bumping the default value of *2 to a larger number or switching this to be a pre-set value of 200 (which is max for 1 page) and then keeping `--limit` as a post-filter cap.

On the filter, I'd drop that entirely as it's unreliable, we're using builds as a source of truth here, filtering them by state skews results entirely.

cmd/job/list.go
scadu: Indeed, `--no-limit` would make more sense here.

JoeColeman95: Another point, 5k results is likely past the max results a user could actually get, so this'd return nothing for most users as they'd get throttled before the query finishes.

JoeColeman95: 5k was design choice, as some may have higher limits/uncapped and it was more of a defensive measure to prevent DoS, rather than something we should aim for.

PR #589, opened 2025-12-10
Migrate Api command and Version commands to Kong

JoeColeman95:
### Description

This PR migrates Api command and Version commands to Kong.

With the version changes, it merges the internal into the cmd, as this doesn't seem like a unnecessary detach. Also, to further this I've removed the import everywhere as it's passed to factory.New(), this can just be imported into the factory itself and then passed as a value directly to the command, rather than passed in at every single use of the func.

Removed internal api.go too, as this wasn't used anywhere. 
### Changes

Moves api to /cmd/api/api.go
Moves version to /cmd/version/version.go
Removes all imports of version except in factory.go
Removes api.go internal, as it wasn't used anywhere 
Moved the internal version to the cmd and then updated goreleaser to update the var directly here instead (tested as working) 

### Testing
- [ x ] Tests have run locally (with `go test ./...`)
- [ x ] Code is formatted (with `go fmt ./...`)

### Disclosures / Credits

Had to slightly refactor api.go defaulting to GET here as it was using Cobra functionality in State(), functionally the same but should be considered.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -1,5 +1,0 @@
-package api_types
-
-type RESTResponse[T any] struct {
-	Data []T `json:"data"`
-}
--- a/pkg/cmd/factory/factory.go
+++ b/pkg/cmd/factory/factory.go
@@ -5,8 +5,8 @@ import (
 	"net/http"
 
 	"github.com/Khan/genqlient/graphql"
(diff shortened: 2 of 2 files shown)


mcncl (approved):
Some nice refactors too (no need for the version passing for a new factory)

PR #597, opened 2025-12-15
feat: add org list command

mcncl:
### Description

When reviewing the `gh` CLI I came across this, which I thought might be a nice add (especially with the `use` command); https://cli.github.com/manual/gh_org_list

### Changes

- Adds a new `organizations` command group, with `list` available
- Aliases to `org`
- Adds aliasing for other `list` commands to `ls`
- Outputs as JSON by default, with a `selected` attribute which changes as we use the `use` command

CleanShot 2025-12-15 at 14 17 59

### Testing


PriyaSudip (approved):
Good addition!

PR #602, opened 2025-12-16
Move `configure` cmd to Kong

lizrabuya:
### Description

Migrate `configure` and `configure add` commands to Kong.

Kong does not handle the pattern well for an optional subcommand like `bk configure add`. The approach was to create a default command that is hidden and takes the `bk configure`.

### Changes

* Moved `/pkg/cmd/configure/*` to `/cmd/configure/*` 
* Removed input prompt for organisation slug and refers to the `--org` flag now. This will error if the flag is not provided. 
 
### Testing

### Disclosures / Credits

Consulted Claude on the best approach to implement the optional subcommand in Kong framework.


pkg/cmd/configure/add/add.go
```diff
@@ -2,68 +2,68 @@ package add
...
 	}
 
-	// Get organization slug
-	org, err := promptForInput("Organization slug: ", false)
-	if err != nil {
-		return err
-	}
-	if org == "" {
-		return errors.New("organization slug cannot be empty")
-	}
```
mcncl, line 51: We no longer ask for the org, so `bk configure add` will fail with _no organisation supplied_ or something, did this get tested?

lizrabuya: @mcncl , yes it got tested. I have removed the prompts and it should be provided as a flag now `--org`. It will error if the flag is not provided. I've updated the PR description. Maybe I should remove it as a flag and change it to an argument. 🤔

mcncl: @lizrabuya it's fine if we want to go via `--org/organization`, however the `--help` output says to run this which is incorrect;

```
bk configure add
```

mcncl: Perhaps we should note in the help output examples that we need those?

mcncl (approved):

PR #607, opened 2025-12-23
Simplify text outputs, remove Lipgloss, BubbleTea/Bubbles

JoeColeman95:
### Description

This PR aims to remove BubbleTea alternate screens, remove lipgloss styling and implement a unicode progress bar for bulk operations. The aforementioned changes have mostly been replaced with a simple tabular view, using ANSI codes for additional styling to improve readability.

All changes have been implemented keeping in mind the variance of different terminals. As such, the new styling (dim, bold, italic, underlined and strikethrough), all respect constraints of terminals, if it's got TERM=dumb, or NO_COLOR, etc. For example, if a terminal is only able to display dim/bold, the text will be formatted as plain-text for other style types due to the ANSI, it will not show the ANSI as an output (tested this with quite a few terminal types)

Further to this, the table in TTY will scale according to the available space, truncating when it would result in line-wrap (as that'd be really ugly). When in non-TTY, this has a fixed width.

When in non-TTY, all formatting is entirely removed, including color and a fixed terminal width of `120` is used. This can be adjusted with `BUILDKITE_TABLE_MAX_WIDTH` (this will be documented in Buildkite Docs after the fact).

The following environment variables are accepted for conforming to standard practises: `NO_PAGER`, `BUILDKITE_NO_PAGER`, `NO_COLOR`, `TERM=dumb`, `PAGER=less`, etc. These will also be documented.

Added a pager across all of the *show me something* items in the CLI, too. This is pretty standard across CLI tools, so it's nice to be part of the club.

Added the availability of `no_pager: bool` to the config (.bk.yaml) to disable this functionality.

`--no-pager` was also added as a global flag to disable the pager (when not auto-detected, or set via env var or config)

Past that, we have a progress bar on `bk agent stop` now, which is nice. I would've done the same for `bk artifacts download`, but that was a major refactor and would've been out of scope.

We still default to JSON everywhere, so this will be invisible for the most part, anyway. These changes are specific to `--output text`

### Changes

- Removed BubbleTea, Lipgloss and Bubbles dependencies
- Deleted `internal/ui` package and all TUI components
- Deleted `internal/agent/*` (entire agent TUI system)
- Deleted `internal/keys/*` and `internal/list/*` (TUI key handlers)
- Removed `pkg/style/style.go` (Lipgloss styling)
- Added pager helper (`internal/io/pager.go`) with support for `PAGER` env var
- Added `--no-pager` global flag to disable paging
- Added `BUILDKITE_NO_PAGER` and `NO_PAGER` environment variables to disable paging
- Added `no_pager: bool` config option in `.bk.yaml`
- Added table renderer (`pkg/output/table.go`) with ANSI styling
- Added `pkg/output/color.go` for color detection (`NO_COLOR`, `TERM=dumb`)
- Added `pkg/output/value.go` for value formatting
- Added `internal/io/progress.go` for progress bar rendering
- Added `BUILDKITE_TABLE_MAX_WIDTH` environment variable to override table width
- Replaced output across all `list`, `view` and `watch` commands to use new table renderer
- Tables auto-size on TTY and use fixed width (default 120) on non-TTY
- ANSI styling respects `NO_COLOR`, `TERM=dumb`, and non-TTY environments
- Added Unicode progress bar to `agent stop` for bulk operations
- Refactored `agent stop` to use worker pool pattern instead of semaphore
- Added tests for all affected areas
 
### Testing

### Disclosures / Credits

- Used Claude when I was getting conflicting info between docs for TTY/non-TTY, color handling, pagers, etc.
- Created some test scaffolding, too, because tests are fun. But had to re-write most of them myself as I kept polishing the codebase after the fact (this can be seen in the commits)
- 100% of the code committed besides tests were written by me (and even then, tests are 90%+ myself)
- 100% of design choices were made by myself
- This has been tested across a number of terminals, ensuring all functionality works as it did before
- No changes are breaking for any flows users may currently have now

### DIFF
--- a/pkg/output/table.go
+++ b/pkg/output/table.go
@@ -275,21 +275,9 @@ func clampColumnWidths(colWidths []int, colCount, separatorWidth, maxWidth int) 
 			clamped[i] = effectiveMin
 			continue
 		}
-
 		ratio := float64(width) / float64(flexSum)
 		alloc := int(math.Floor(ratio * float64(remainingSpace)))
-
-		alloc = min(alloc, effectiveMin)
-
-		if width <= narrowThreshold {
-			// This column is narrow enough to get its full width
-			clamped[i] = width
-			fixed[i] = true
-			remainingSpace -= width
-		} else {
-			// This column needs proportional allocation
-			flexSum += width
-		}
+		clamped[i] = max(alloc, effectiveMin)
 	}
 
 	currentTotal := 0


pkg/output/table.go
```diff
@@ -0,0 +1,324 @@
...
+	if available < minColumnWidth*colCount {
+		effectiveMin = available / colCount
+	}
+
+	for i, width := range colWidths {
+		ratio := float64(width) / float64(sum)
+		alloc := int(math.Floor(ratio * float64(available)))
+		if alloc < effectiveMin {
+			alloc = effectiveMin
+		}
```
mcncl, line 277: As far as I can tell, where there is a significantly long value in the second column, like a URL, long message, description etc, we would truncate the left hand column to be something like this for `agent view ...`

```sh
P...
```

```suggestion
    // First pass: give narrow columns their full width, mark wide columns for proportional allocation
    // A column is "narrow" if it fits within its fair share of space
    fairShare := available / colCount
    narrowThreshold := fairShare * 2
    
    fixed := make([]bool, len(colWidths))
    remainingSpace := available
    flexSum := 0

	for i, width := range colWidths {
	    if width <= narrowThreshold {                                                                                                                                    
	      // This column is narrow enough to get its full width
	      clamped[i] = width
	      fixed[i] = true
          remainingSpace -= width
        } else {
          // This column needs proportional allocation
          flexSum += width
        }
    }

    // Second pass: proportionally allocate remaining space to flexible columns
    for i, width := range colWidths {
        if fixed[i] {
            continue
        }
    if flexSum == 0 {
        clamped[i] = effectiveMin
        continue
    }
    
    ratio := float64(width) / float64(flexSum)
    alloc := int(math.Floor(ratio * float64(remainingSpace)))

    if alloc < effectiveMin {
		alloc = effectiveMin
	}
```

mcncl (approved):

PR #623, opened 2026-01-15
Add `config` command for set/get

mcncl:
## Description

This adds a `config` command which allows users to `set`/`get`/`unset`/`list`. This enables folks to set config options via the CLI instead of having to edit the config file directly.

## Changes

- adds a `config` command in `cmd/` as well as tests for it
- adds setters and getters for the `config`
- adds config to the `main.go`
- adds tests for the internal package

## Testing

## Caveats

I've added an alias to `configure` so that folks can use `auth`, too. Having `configure` and `config` may be confusing and I don't think `configure` is the correct verb for the act of adding org/token.

The alias will flip once we move to use `auth` as the command instead, and will be retired eventually (next major bump).

## Disclosures / Credits

I used Claude Code to plan out these changes as well as for back-and-forth on the best approach/precedence of the options. Claude was also used to write the `*_test.go` files to ensure we have adequet coverage, though it didn't initially cover all cases, so I added those.

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,7 +98,7 @@ type (
 		Copy     pipeline.CopyCmd     `cmd:"" help:"Copy an existing pipeline." aliases:"cp"`
 		Create   pipeline.CreateCmd   `cmd:"" help:"Create a new pipeline."`
 		List     pipeline.ListCmd     `cmd:"" help:"List pipelines." aliases:"ls"`
-		Migrate  pipeline.MigrateCmd  `cmd:"" help:"Migrate a CI/CD pipeline configuration to Buildkite format."`
+		Convert  pipeline.ConvertCmd  `cmd:"" help:"Convert a CI/CD pipeline configuration to Buildkite format." aliases:"migrate"`
 		Validate pipeline.ValidateCmd `cmd:"" help:"Validate a pipeline YAML file."`
 		View     pipeline.ViewCmd     `cmd:"" help:"View a pipeline."`
 	}
--- a/cmd/pipeline/convert.go
+++ b/cmd/pipeline/convert.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
@@ -18,15 +19,15 @@ import (
 	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 )
 
-var MigrationEndpoint = "https://m4vrh5pvtd.execute-api.us-east-1.amazonaws.com/production/migrate"
+var convertEndpoint = "https://m4vrh5pvtd.execute-api.us-east-1.amazonaws.com/production/migrate"
 
-type migrationRequest struct {
+type conversionRequest struct {
 	Vendor string `json:"vendor"`
 	Code   string `json:"code"`
 	AI     bool   `json:"ai,omitempty"`
 }
 
-type migrationResponse struct {
+type conversionResponse struct {
 	JobID     string `json:"jobId"`
 	Status    string `json:"status"`
 	Message   string `json:"message"`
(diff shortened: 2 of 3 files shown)


cmd/agent/list.go
```diff
@@ -32,7 +32,7 @@ type ListCmd struct {
 	Tags     []string `help:"Filter agents by tags"`
 	PerPage  int      `help:"Number of agents per page" default:"30"`
 	Limit    int      `help:"Maximum number of agents to return" default:"100"`
-	Output   string   `help:"Output format. One of: json, yaml, text" short:"o" default:"${output_default_format}" enum:"json,yaml,text"`
+	Output   string   `help:"Output format. One of: json, yaml, text" short:"o" default:"${output_default_format}" enum:",json,yaml,text"`
```
scadu, line 35:
```suggestion
	Output   string   `help:"Output format. One of: json, yaml, text" short:"o" default:"${output_default_format}" enum:"json,yaml,text"`
```

Unless the comma was intentional here? If so, why? It's done in other places, too.

main.go
```diff
@@ -56,6 +57,12 @@ type (
 	VersionCmd struct {
 		version.VersionCmd `cmd:"" help:"Print the version of the CLI being used"`
 	}
+	ConfigCmd struct {
```
scadu: Initially, it's not clear to me how values for each setting should be passed. We might want to include a brief example, like `bk config set output_format text`.

Also, a list of all the available options and their possible values, not only ones already set in the config, would be super handy.

cmd/config/set.go
```diff
@@ -0,0 +1,69 @@
...
+	// Determine where to save (default to user config unless --local)
+	saveLocal := c.Local && inGitRepo
+
+	switch key {
+	case KeySelectedOrg:
+		return conf.SelectOrganization(c.Value, saveLocal)
+	case KeyOutputFormat:
+		return conf.SetOutputFormat(c.Value, saveLocal)
+	case KeyNoPager:
+		v, _ := strconv.ParseBool(c.Value)
```
scadu: Shall we wrap it in `err`?
```suggestion
		v, err := strconv.ParseBool(c.Value)
		if err != nil {
    		return fmt.Errorf("invalid boolean value %q: %w", c.Value, err)
		}
```

cmd/config/set.go
```diff
@@ -0,0 +1,69 @@
...
+	if c.Local && !inGitRepo {
+		return fmt.Errorf("--local requires being in a git repository")
+	}
+
+	// Determine where to save (default to user config unless --local)
+	saveLocal := c.Local && inGitRepo
+
+	switch key {
+	case KeySelectedOrg:
+		return conf.SelectOrganization(c.Value, saveLocal)
```
scadu: Can we extract it to some common function, like `validateAndSaveConfig`?

internal/config/config.go
```diff
@@ -172,6 +176,93 @@ func (conf *Config) PagerDisabled() bool {
...
 }
 
+// SetNoPager sets whether the pager is disabled
+func (conf *Config) SetNoPager(v bool, inGitRepo bool) error {
+	if !inGitRepo {
+		conf.user.NoPager = v
+		return conf.writeUser()
+	}
+	conf.local.NoPager = v
+	return conf.writeLocal()
```
scadu: The `!inGitRepo` feels backwards to me. Shouldn't we start from the local-scoped settings?

```suggestion
	if inGitRepo {
    	conf.local.NoPager = v
    	return conf.writeLocal()
	}
	conf.user.NoPager = v
	return conf.writeUser()
```

cmd/agent/list.go
mcncl: The comma here means that Kong can accept an empty string/no value for the `output` flag, then in our `ResolveFormat` we move through the conditionals to see what to do if `flagValue == ""` (we check the configValue), else we use the default.

Without it we'd fall back to default where no value is used and not be able to use the config value.

main.go
mcncl: I'd be OK with adding a `bk config list` with output like `git config --list` has

cmd/config/set.go
mcncl: We can, but at this point we've already validated in;

```go
if validValues := key.ValidValues(); validValues != nil
```

Happy to commit your suggestion, though

internal/config/config.go
mcncl: I think the logic is correct here but the naming could be confusing; perhaps `inGitRepo` should be `saveToLocal`

cmd/agent/list.go
mcncl: `enum:",json,yaml,text"` would resolve to `"",json,yaml,text`

scadu: Ahh, that’s what I suspected but wasn’t sure. Thanks for clarifying!

main.go
mcncl: I've added verbose help text for each command

scadu (approved):

PR #626, opened 2026-01-18
add more protection to releases

mcncl:
### Description

Creating a release is a little risky right now as we allow for any tag, so a person may skip a `minor` version, set the incorrect _type_ of tag or even release a `major` version.

### Changes

- updates the `tag.sh` to check if a major version bump has been entered and errors out to inform the user that a major bump should be done manually
- changes the `block` step to an arguably more useful `input` step which;
	- asks the user to _select_ what type of tag they wish to use (major[see above]/minor/patch)
	- sets teams which are allowed to interact with this step
	- sets a `blocked_state` to replicate current `blocked` behaviour (build passing)

### DIFF
--- a/cmd/job/log.go
+++ b/cmd/job/log.go
@@ -17,9 +17,9 @@ import (
 
 type LogCmd struct {
 	JobID        string `arg:"" help:"Job UUID to get logs for"`
-	Pipeline     string `help:"The pipeline to use. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}." short:"p"`
-	BuildNumber  string `help:"The build number." short:"b"`
-	NoTimestamps bool   `help:"Strip timestamp prefixes from log output." name:"no-timestamps"`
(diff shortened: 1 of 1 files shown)


.buildkite/tag.sh
```diff
@@ -1,12 +1,42 @@
...
+RELEASE_TYPE="$(buildkite-agent meta-data get "release-type")"
+
+if [[ "${RELEASE_TYPE}" == "major" ]]; then
+  echo "🚨 Major releases require manual tagging to prevent accidents."
+  echo "Please run: git tag vX.0.0 && git push origin vX.0.0"
+  exit 1
+fi
+
+# Get latest tag matching v*.*.* pattern
+LATEST_TAG=$(git describe --tags --match "v[0-9]*" --abbrev=0 2>/dev/null || echo "v3.0.0")
```
scadu: I'd rather have the process fail right away, then use `v3.0.0` value as a fallback and fail just before creating a new tag.

PR #637, opened 2026-02-05
Add initial OAuth support

mcncl:
### Description

Implements OAuth support via `bk auth login`.

### Changes

- Adds `OAUTH_CLIENT_ID` to pipeline (and in Cluster secrets)
	- It's not strictly private, but this is best practise
- Updates `.goreleaser` to support injection of OAuth Client ID in built binaries
- Adds `login` and `logout` subcommands to `auth` root
	- `bk auth login` & `bk auth logout`
- Stores tokens in keyring (OS)
- Removes `auth` alias for `configure` (that was temporary anyway)
- Adds some basic OAuth function; authorized page etc in browser

### Testing

### Caveats

Still to come; removing tokens from text-based config.

### Disclosures / Credits

I had used Claude to implement the OAuth endpoint in bk/bk due to my lack of Rails experience, so I consulted with it to ensure the implementation approach for the CLI would be in line, functionally, with how the bk/bk OAuth was built.

### DIFF
--- a/pkg/keyring/keyring.go
+++ b/pkg/keyring/keyring.go
@@ -64,7 +64,7 @@ func (k *Keyring) IsAvailable() bool {
 
 // MockForTesting replaces the keyring backend with an in-memory store
 // and resets the availability cache so subsequent New() calls reflect
-// the mock. 
+// the mock.
 func MockForTesting() {
 	keyring.MockInit()
 	keyringAvailableOnce = sync.Once{}


pkg/oauth/oauth.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate state: %w", err)
+	}
+
+	var listener net.Listener
+
+	// Only start local callback server if no custom redirect URI provided
+	if cfg.CallbackURL == "" {
+		var err error
+		listener, err = net.Listen("tcp", "localhost:0")
```
scadu: As per [RFC 8252](https://www.rfc-editor.org/rfc/rfc8252), we should listen on `127.0.0.1`, not `localhost`

JoeColeman95 (approved):

JoeColeman95:
Looks great 🔒

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,286 @@
...
+	if f.config.OrgUUID != "" {
+		params.Set("organization_uuid", f.config.OrgUUID)
+	}
+
+	return fmt.Sprintf("https://%s/oauth/authorize?%s", f.config.Host, params.Encode())
+}
+
+// WaitForCallback waits for the OAuth callback and returns the authorization code
+func (f *Flow) WaitForCallback(ctx context.Context) (*CallbackResult, error) {
+	resultCh := make(chan *CallbackResult, 1)
```
JoeColeman95, line 139:
```suggestion
	if f.listener == nil {
		return nil, fmt.Errorf("callback listener not available as a custom CallbackURL was provided")
	}

	resultCh := make(chan *CallbackResult, 1)
```

edge case to handle panic

cmd/auth/login.go
```diff
@@ -0,0 +1,121 @@
...
+	// Check if already configured
+	existingToken := f.Config.GetTokenForOrg(c.Org)
+	if existingToken != "" && !globals.SkipConfirmation() {
+		fmt.Printf("Organization %q is already configured. Use 'bk configure --force' to override.\n", c.Org)
+		return nil
+	}
+
+	// Create OAuth flow
+	cfg := &oauth.Config{
+		Host:     oauth.DefaultHost,
```
JoeColeman95:
```suggestion
		// Host default handled via NewFlow, omitted to allow usage of BUILDKITE_HOST
```

Would remove host here or set it to an empty string so that `NewFlow` can apply `BUILDKITE_HOST` override, as it also does `DefaultHost` there (buildkite.com const) as a fallback, otherwise `BUILDKITE_HOST` will always be skipped due to host always being buildkite.com

pkg/keyring/keyring.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Test if keyring is functional by attempting a dummy operation
+	testKey := "buildkite-cli-keyring-test"
+	err := keyring.Set(serviceName, testKey, "test")
+	if err != nil {
+		return false
+	}
+	_ = keyring.Delete(serviceName, testKey)
+	return true
+}
```
JoeColeman95, line 97: Two things here, probably a bit nit-picky.

UX: (expiration of keychain access will cause this) whenever invocation occurs here it could potentially prompt for keychain access down the line when it's not actually needed due to the token write, this could probably be optimised with sync.Once to cache the bool that you're returning as a var, meaning the value will be returned without having to write multiple times. Just my 2¢, not going to block on this one if you feel this is a reasonable default. It may well be better to do it this way to remain secure (user locks their machine and comes back, they'll have to reauth keychain), but the call every time is negligible. If this is designed as intended, fair enough, just resolve this one.

nit: just my lack of trust of technology, but I'd do a rand string or uuid on the testKey so if the delete fails (super edgecase, IK), it won't be breaking the next time around. That, or an error message that would prompt the user to manually intervene, but the former's probably best for UX IMO. Again, feel free to resolve this convo too.

JoeColeman95 (dismissed):
Approving as last outstanding is not breaking and can be iterated on if needed

PR #638, opened 2026-02-09
Add commands for secrets

mcncl:
### Description

Adds support for `secrets` use; create, update, delete, list and get

### Changes

- adds a command for `bk secret`, allowing the user to `create`, `update`, `delete,` `list`, `get`
- includes testing for the secrets API

### Testing

### Disclosures / Credits

Sure, I'm a modern person. I worked with AI to plan this work, it also highlighted some issues with the `go-buildkite` API which have been fixed/implemented as part of that.

The future is now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/buildkite/cli/v3
 go 1.25
 
 require (
-	github.com/alecthomas/kong v1.13.0
+	github.com/alecthomas/kong v1.14.0
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be
 	github.com/buildkite/go-buildkite/v4 v4.14.1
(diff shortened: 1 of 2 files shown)


cmd/secret/create.go
```diff
@@ -0,0 +1,112 @@
...
+
+If --value is not provided, you will be prompted to enter the secret value
+interactively (input will be masked).
+
+Examples:
+  # Create a secret with interactive value input
+  $ bk secret create --cluster-id my-cluster-id --key MY_SECRET
+
+  # Create a secret with the value provided inline
+  $ bk secret create --cluster-id my-cluster-id --key MY_SECRET --value "s3cr3t"
```
omehegan, line 40: Do we just trust people to be sophisticated enough to be aware of the risks of this?

PR #649, opened 2026-02-16
pipeline convert help corrections

karensawrey:
### Changes

This PR changes nothing in the way CLI fuctions, but changes help descriptions for auto-detection on vendors and the use of timeouts.
-->

### Testing
Does not apply.

### Disclosures / Credits

Minor fix in the descriptions.
-->

### DIFF
--- a/cmd/auth/login.go
+++ b/cmd/auth/login.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -17,6 +18,8 @@ import (
 
 type LoginCmd struct {
 	Scopes string `help:"OAuth scopes to request" default:""`
+	Org    string `help:"Organization slug (required with --token)" optional:""`
+	Token  string `help:"API token to store (non-OAuth login)" optional:""`
 }
 
 func (c *LoginCmd) Help() string {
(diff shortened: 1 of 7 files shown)


cmd/pipeline/convert.go
```diff
@@ -45,9 +45,9 @@ type statusResponse struct {
 
 type ConvertCmd struct {
 	File    string `help:"Path to the pipeline file to convert (required)" short:"F" required:""`
-	Vendor  string `help:"CI/CD vendor (auto-detected if not specified)" short:"v"`
+	Vendor  string `help:"CI/CD vendor (auto-detected if the file name matches vendor path and name - otherwise, needs to be specified)" short:"v"`
 	Output  string `help:"Custom path to save the converted pipeline (default: .buildkite/pipeline.<vendor>.yml)" short:"o"`
-	Timeout int    `help:"Timeout in seconds for conversion" default:"300"`
+	Timeout int    `help:"Timeout in seconds for conversion (increase for large pipelines, decrease to fail conversion on timeout)" default:"300"`
```
scadu: I'm not sure if it's clear to me what's the intent of the current wording, especially around decreasing the timeout value.

karensawrey: @omehegan we might need your assistance here as I tried to convey your explanation to the best of my ability, and it didn't seem to have worked.

mcncl: Yeah, I don't know that this needs a change, maybe I'm biased here as I know what `timeout` is, but I think it's generally accepted as a term and function?

`decrease to fail conversion on timeout`, what happens if I set to `0`?

karensawrey: @mcncl if the timeout is set to zero, same thing happens as when the timeout is shorter than the conversion needs to run - "panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x104d337d0]"

karensawrey: Should we add a warning to advise users against setting the timeout to zero, to be on the safe side?

mcncl: No, that should be handled in the code

mcncl: Fixed in https://github.com/buildkite/cli/pull/655

karensawrey: Thanks, @mcncl! 

Hey, @omehegan could you please take a look here and help us out with a better explanation of time-outs?

omehegan: I do find the "decrease to fail conversion on timeout" to be slightly hard to parse. The intent is simply this: if you are converting large pipelines, you might need to increase this value. If you are converting small ones, you _could_ decrease the value so that it fails faster if something goes wrong. I think in practice no one will look at this unless they hit a timeout, so maybe we just remove the "increase if/decrease to" wording and leave the rest. It's a timeout, people will adjust it if they need to.

mcncl:
```suggestion
	Timeout int    `help:"The time (in seconds) after which a conversion should be cancelled" default:"300"`
```

omehegan (dismissed):

PR #702, opened 2026-03-19
ci: switch builds from docker compose to mise

lox:
## Summary
- replace the remaining docker-compose-based Buildkite steps with the mise plugin
- enable hosted-agent cache volumes for mise and point Go/golangci-lint caches at that volume
- lock the shared toolchain in mise, including goreleaser-pro and ko, and remove the obsolete Buildkite Docker files

## Testing
- go run . pipeline validate --file .buildkite/pipeline.yml --file .buildkite/pipeline.release.yml
- git diff --check
- mise install
- mise exec -- goreleaser --version
- mise exec -- ko version


.buildkite/pipeline.release.yml
```diff
@@ -12,49 +21,32 @@ steps:
     secrets:
       - POSTHOG_API_KEY
       - OAUTH_CLIENT_ID
+    env:
+      <<: *cache_env
+      GOOS: "{{matrix}}"
+    commands:
+      - '. "${BUILDKITE_ENV_FILE}"'
```
lucaswilric: I'm not sure I understand why it's necessary to set env vars by appending them to `BUILDKITE_ENV_FILE` and then having an extra command that's required in order for the plugin to work.

Is there a reason not to set the vars directly ([example](https://github.com/buildkite/tool-versions-env-buildkite-plugin/blob/main/hooks/post-checkout#L21))? My gut says that would separate the `mise` concern more neatly, but maybe there's something I don't know?

lox: Yup, agreed, fixing this upstream in the mise plugin.

.buildkite/pipeline.release.yml
```diff
@@ -1,6 +1,11 @@
 agents:
   queue: hosted
 
+x-mise-plugin: &mise_plugin
+  buildkite-plugins/mise#v1.1.1: ~
```
catkins:
```suggestion
  mise#v1.1.1: ~
```

.buildkite/pipeline.yml
```diff
@@ -1,28 +1,27 @@
 agents:
   queue: hosted
 
+x-mise-plugin: &mise_plugin
+  buildkite-plugins/mise#v1.1.1: ~
```
catkins:
```suggestion
  mise#v1.1.1: ~
```

.buildkite/pipeline.release.yml
lox: Fixed in a145530: updated the release pipeline to `buildkite-plugins/mise#v1.1.1` once the upstream plugin fix landed.

.buildkite/pipeline.yml
lox: Fixed in a145530: updated the main pipeline to `buildkite-plugins/mise#v1.1.1` once the upstream plugin fix landed.

.buildkite/pipeline.yml
```diff
@@ -1,28 +1,35 @@
 agents:
   queue: hosted
 
+x-cache-env: &cache_env
+  GOCACHE: .buildkite/cache-volume/go/build
```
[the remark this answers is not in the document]
lox: Fixed in b376e2b: the cache exports now come from a checkout-rooted hook, so Go and golangci-lint get absolute cache paths in both the main and release pipelines.

.buildkite/hooks/pre-command
```diff
@@ -12,3 +12,14 @@ mkdir -p \
 export GOCACHE="${cache_root}/go/build"
 export GOMODCACHE="${cache_root}/go/pkg/mod"
 export GOLANGCI_LINT_CACHE="${cache_root}/golangci-lint"
+
+# CI pins tool versions in mise.toml/mise.lock, but the go: backend used for
+# go-cover-treemap does not support locked-mode URLs. Skip the GitHub API-backed
```
catkins: Could we provide a readonly MISE_GITHUB_TOKEN with only public repo access to bypass the rate limits?

.buildkite/pipeline.yml
lox: Finally actually fixed this.

lox:
@catkins mind stamping this again? Finally fixed the short plugin reference that you mentioned several times.

.buildkite/pipeline.yml
```diff
@@ -1,35 +1,37 @@
...
-          run: golangci-lint
-          tty: true
+      - mise#v1.1.1: ~
 
   - name: ":go: test"
     key: test
     artifact_paths:
       - cover-tree.svg
+    secrets:
+      - MISE_GITHUB_TOKEN
```
catkins, line 20: Is this used in a public pipeline? If so we may want to use a different token with only access to public stuff. 

EDIT: nvm that's already one! Thanks.

catkins (approved):

PR #727, opened 2026-03-29
[FEAT]: Add `team` command group

mcncl:
### Description

Adds support for group of `team` commands, as available in [go-buildkite](https://github.com/buildkite/go-buildkite/blob/main/teams.go)

### Changes

- adds `team {create,delete,list,update,view}`
- adds tests for the new commands
- updates `main.go` to support

```sh
❯ go run . team create Testers
{
  "id": "89cea5f7-a7c0-4183-b48c",
  "name": "Testers",
  "slug": "testers",
  "privacy": "visible",
  "default": false,
  "created_at": "2026-03-29T23:20:04.042Z",
  "created_by": {
    "id": "01831b25-7d66-431e-8dcf",
    "name": "Ben McNicholl",
    "email": "ben.mcnicholl@buildkite.com",
    "created_at": "2022-09-08T03:29:45.318Z"
  }
}

❯ go run . team list
[
  {
    "id": "3c911fd5-cf8f-4a45-814d",
    "name": "Everybody",
    "slug": "everybody",
    "description": "The default team",
    "privacy": "visible",
    "default": true,
    "created_at": "2026-03-16T02:08:21.763Z",
    "created_by": {
      "id": "01831b25-7d66-431e-8dcf",
      "name": "Ben McNicholl",
      "email": "ben.mcnicholl@buildkite.com",
      "created_at": "2022-09-08T03:29:45.318Z"
    }
  },
  {
    "id": "89cea5f7-a7c0-4183-b48c",
    "name": "Testers",
    "slug": "testers",
    "privacy": "visible",
    "default": false,
    "created_at": "2026-03-29T23:20:04.042Z",
    "created_by": {
      "id": "01831b25-7d66-431e-8dcf",
      "name": "Ben McNicholl",
      "email": "ben.mcnicholl@buildkite.com",
      "created_at": "2022-09-08T03:29:45.318Z"
    }
  }
]

❯ go run . team delete 89cea5f7-a7c0-4183-b48c --yes
Team deleted successfully.
```

### Testing

### DIFF
--- a/cmd/team/list.go
+++ b/cmd/team/list.go
@@ -1,0 +1,145 @@
+package team
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/alecthomas/kong"
+	"github.com/buildkite/cli/v3/internal/cli"
+	bkIO "github.com/buildkite/cli/v3/internal/io"
+	"github.com/buildkite/cli/v3/internal/team"
+	"github.com/buildkite/cli/v3/pkg/cmd/factory"
+	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	"github.com/buildkite/cli/v3/pkg/output"
+	buildkite "github.com/buildkite/go-buildkite/v4"
+)
+
+type ListCmd struct {
+	PerPage int `help:"Number of teams per page" default:"30"`
+	Limit   int `help:"Maximum number of teams to return" default:"100"`
+	output.OutputFlags
+}
+
+func (c *ListCmd) Help() string {
+	return `
+List the teams for an organization. By default, shows up to 100 teams.
+
+Examples:
+  # List all teams
+  $ bk team list
+
+  # List teams in JSON format
+  $ bk team list -o json
+
+  # List up to 200 teams
+  $ bk team list --limit 200
+`
+}
+
+func (c *ListCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
+	f, err := factory.New(factory.WithDebug(globals.EnableDebug()))
+	if err != nil {
+		return err
+	}
+
+	f.SkipConfirm = globals.SkipConfirmation()
+	f.NoInput = globals.DisableInput()
+	f.Quiet = globals.IsQuiet()
+	f.NoPager = f.NoPager || globals.DisablePager()
+
+	if err := validation.ValidateConfiguration(f.Config, kongCtx.Command()); err != nil {
+		return err
+	}
+
+	format := output.ResolveFormat(c.Output, f.Config.OutputFormat())
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	teams, hasMore, err := listTeams(ctx, f, c.PerPage, c.Limit)
+	if err != nil {
+		return err
+	}
+
+	if format != output.FormatText {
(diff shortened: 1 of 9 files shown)


cmd/team/list.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return nil
+}
+
+func listTeams(ctx context.Context, f *factory.Factory) ([]buildkite.Team, error) {
+	var teams []buildkite.Team
+	var err error
+
+	spinErr := bkIO.SpinWhile(f, "Loading teams information", func() {
+		teams, _, err = f.RestAPIClient.Teams.List(ctx, f.Config.OrganizationSlug(), nil)
```
omehegan: Any concerns about not paginating here? I'm not sure how many teams we return per page, but I'm guessing sooner or later we will come across someone who has hundreds of them.

cmd/team/update.go
```diff
@@ -0,0 +1,142 @@
...
+		return fmt.Errorf("error fetching team: %v", err)
+	}
+
+	// Build update input from current values, overriding with any flags set
+	input := buildkite.CreateTeam{
+		Name:                      current.Name,
+		Description:               current.Description,
+		Privacy:                   current.Privacy,
+		IsDefaultTeam:             current.Default,
+		MembersCanCreatePipelines: false,
```
omehegan: Claude flagged the fact that every time someone updates a team, it's going to set `members_can_create_pipelines: false` even if the current setting for that team is `true`. "Balderdash," said I. "Prove it!" So we looked at go_buildkite, where it turns out that the Team struct doesn't include all the fields that the API returns, including `MembersCanCreatePipelines`. And then looking upstream at the API in `Team::Updater` and how it would handle this, it looks like it would in fact apply the `false` flag and change the setting unexpectedly. We'd need to fix this in go_buildkite and then here do `MembersCanCreatePipelines: current.MembersCanCreatePipelines,`

If this makes sense to you, I can throw the PR at go_buildkite and then add a commit to this.

omehegan (dismissed):

cmd/team/update.go
mcncl: I think `false` is a sensible default? It's definitely safer than `true` and would still allow for creation of the team but with safeguards in place.

omehegan: Discussed in Slack - we still need to fix go_buildkite to avoid us forcibly changing this setting when we update teams. That'll be a separate task.

cmd/team/list.go
```diff
@@ -0,0 +1,159 @@
...
+		if page > 1 && pageTeams[0].ID == previousFirstTeamID {
+			return nil, false, fmt.Errorf("API returned duplicate page content at page %d, stopping pagination to prevent infinite loop", page)
+		}
+		previousFirstTeamID = pageTeams[0].ID
+
+		all = append(all, pageTeams...)
+
+		if len(all) > limit {
+			hasMore = true
+		}
```
scadu: Minor: Could we only set `hasMore` after confirming another result exists? When the total is exactly equal to `--limit` and the final page is full, this reports `Showing 2+ teams` even though page 2 is empty. Fetching one extra result or page would distinguish an exact boundary from truncated output. A test where the total equals both `--limit` and `--per-page` should cover it.

omehegan:
@scadu since this PR had gone a little stale, Ben has changed teams, and is now off for a month, I decided to try to land it. Made a bunch of changes in response to Buildsworth feedback. Could you give this a final review and merge if happy?

scadu (approved):

PR #737, opened 2026-03-31
preflight: if the repo has no remote, skip push & cleanup

pda:
When a git repo has no remote, preflight would previously do this:

```
Error: snapshot error: git push origin 3f6b6eaf03a0d2eaf7be4a02ede2ab34740b29cc:refs/heads/bk/preflight/019d4313-4222-730d-9669-98483dc2e507: exit status 128 (failed to create preflight snapshot)
```

There are useful workflows where a buildkite-agent runs on the same machine as a git repo, with the pipeline configured with a repository of `/path/to/codebase/.git`. Changes to the that git repo don't need to be pushed anywhere for the agent/build/preflight to use them.

Emit a warning in this case, because it could also be an accidental misconfiguration.


mcncl (approved):

pda:
I'll close this for now as it's pretty merge-conflicty, might revisit it next time I run into it.

PR #773, opened 2026-04-09
Add OAuth refresh token support

catkins:
### Description

When the Buildkite API issues short-lived access tokens with refresh tokens during OAuth login, the CLI currently has no way to renew them — every expired token results in a 401 and a manual `bk auth login`.

This adds transparent refresh token support: expired access tokens are automatically renewed via the OAuth refresh grant without user intervention.

### Design

- **Shared `TokenSource`**: A mutable token holder read by `AuthTransport` on every request (both REST and GraphQL). After a refresh, all subsequent requests immediately use the new token — no stale cached values.
- **Compare-after-lock guard**: When concurrent requests hit 401, only the first performs the refresh. Others detect the token already changed and skip. Critical with token rotation where reuse revokes the entire family.
- **Selective error handling**: Only terminal OAuth errors (`invalid_grant`, `unauthorized_client`) clear the stored refresh token. Transient failures (network, timeouts) preserve it.
- **Body buffering**: Request bodies are buffered so POST requests (GraphQL mutations) can be replayed on retry.
- **Debug redaction**: Token values are redacted in `--debug` output for both request bodies and responses.

### Testing
- [x] Linting passes (`golangci-lint run`)
- Unit tests for: compare-after-lock, transient error preservation, body buffering, concurrent coordination, terminal vs transient error classification, `TokenSource` thread safety, `RefreshAccessToken` success/error
- Verified full flow against local dev server (auth code → refresh token → token rotation → reuse rejection)

### Disclosures / Credits

This PR was authored with Amp (Claude).

### DIFF
--- a/cmd/auth/login.go
+++ b/cmd/auth/login.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/alecthomas/kong"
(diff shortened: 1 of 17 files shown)


mcncl (approved):
Looks good to me overall once conflicts are resolved, the AI review comment is a non-blocker for this piece of work, but might be easy to just _commit suggestion_.

PR #775, opened 2026-04-10
Adjust http client to support rate limits

mcncl:
### Description

Changed the way that the CLI implements rate limits and the back off. We've removed the dependency on `roko`, which shouldn't be required with the availability of the `RateLimit-` headers that we can get from client calls.

Moves the logic in to a reusable HTTP transport so commands/clients can share it.

### Changes

- Added RateLimitTransport in internal/http/ratelimit.go:
  - Retries on HTTP 429 up to MaxRetries
  - Uses RateLimit-Reset for backoff, with MaxRetryDelay cap
- Removed retry/backoff code from internal/http/client.go:
  - Deleted WithMaxRetries, WithMaxRetryDelay, WithOnRetry
  - Removed RetryAfter/retry loop from Client.Do
- Updated call sites to use transport-based retry:
- Extended pkg/cmd/factory/factory.go with WithTransport(http.RoundTripper) and composed it with debug transport.

#### Caveats

- Removed github.com/buildkite/roko from go.mod/go.sum.

### Testing

### Disclosures / Credits

I used AI assistance (OpenCode / GPT-5.3 Codex) to review the diff, it highlighted a regression in some logic I'd written where if `RateLimit-Reset` was _equal_ to `0` on check it would wait 10s.


cmd/preflight/preflight.go
```diff
@@ -90,6 +92,15 @@ func (c *PreflightCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error
 
 	renderer := newRenderer(os.Stdout, c.JSON, c.Text, stop)
 
+	rlTransport.OnRateLimit = func(attempt int, delay time.Duration) {
+		_ = renderer.Render(Event{
+			Type:        EventOperation,
+			Time:        time.Now(),
+			PreflightID: preflightID.String(),
+			Title:       fmt.Sprintf("Rate limited by API, waiting %s before retrying (attempt %d/%d)...", delay.Truncate(time.Second), attempt+1, rlTransport.MaxRetries),
+		})
```
matthewborden: Should this only be printed in a debug or verbose mode?

mcncl: Hmm, that would mean that the agent would have to know to run in that mode in order to see API feedback. An agent would likely run bk preflight not in --debug and would then just see builds hang

pkg/cmd/factory/factory.go
```diff
@@ -55,6 +56,14 @@ func WithOrgOverride(org string) FactoryOpt {
 	}
 }
 
+// WithTransport sets a custom http.RoundTripper for the REST API client.
+// It is composed with the debug transport when debug mode is enabled.
+func WithTransport(t http.RoundTripper) FactoryOpt {
+	return func(c *factoryConfig) {
+		c.transport = t
+	}
+}
```
matthewborden, line 65: Does this overwrite the retry behaviour for 429s in https://github.com/buildkite/go-buildkite/blob/main/buildkite.go#L383-L403? Could there be two backoff loops on top of each other for 429s if it doesn't?

mcncl: @matthewborden it does; `go-buildkite` uses a generic exponential backoff which will act as a fallback if our checks fail for some reason (missing headers or something)

We're using the `RateLimit-Reset`, which will mean we have an exact time for retrying, rather than the exponential value that `go-buildkite` might use

matthewborden: Let's address this in a followup PR, I'd prefer to have one backoff loop managing rate limit retries.

matthewborden (approved):

PR #780, opened 2026-04-13
Allows for preflight to log job retries and respect their outcome

mcncl:
### Description

Allows for job retries to be logged and respected by preflight, similarly to how test retries can be.

### Changes

- adds an event for job retried and allows preflight to consume that via the watcher
- presents the information to the user (open to feedback on messaging here, this was just to get the PR open)
- allows for the tracker to track retried jobs
- adds the event to the renderer
- expands the testing of the `preflight` package to cover job retrying

### Testing


internal/build/watch/tracker.go
```diff
@@ -159,7 +180,7 @@ func sortJobsByStartTime(jobs []buildkite.Job) {
 func (t *JobTracker) summarize(b buildkite.Build) JobSummary {
 	var s JobSummary
 	for _, j := range b.Jobs {
-		if j.Type != "script" {
+		if j.Type != "script" || j.Retried {
```
matthewborden, line 183: This would include retried trigger jobs.

internal/build/watch/tracker.go
```diff
@@ -106,31 +108,50 @@ func (t *JobTracker) Update(b buildkite.Build) BuildStatus {
 		}
 	}
 
+	// Second pass: detect retry jobs that just reached passed.
+	for _, j := range b.Jobs {
+		if j.Type != "script" || j.State != "passed" || j.RetriesCount == 0 {
```
matthewborden:
```suggestion
		if j.Type != "script" && (j.State != "passed" || j.RetriesCount == 0) {
```

internal/build/watch/tracker.go
mcncl: @matthewborden so you'd like retried trigger jobs to log, too?

matthewborden (approved):

PR #786, opened 2026-04-15
fix: keep diagnostics off stdout for structured output

mtibben:
Support machine-readable CLI output by keeping diagnostics off stdout.

Structured output commands currently print warnings and empty-state messages to stdout, which breaks callers that expect stdout to contain only JSON or YAML. Browser-open failures also use stdout even though they are diagnostics.

This routes shared warnings to stderr and makes JSON- and YAML-capable commands return structured empty values when no resources are found. It also adds focused regression coverage around the shared output helper and configuration validation warning so structured output stays clean.

### DIFF
--- a/pkg/cmd/validation/config_test.go
+++ b/pkg/cmd/validation/config_test.go
@@ -61,11 +61,14 @@ func TestValidateConfiguration_MissingValues(t *testing.T) {
 		t.Setenv("BUILDKITE_ORGANIZATION_SLUG", "")
 		conf := newTestConfig(t)
 
+		var validationErr error
 		stdout, stderr := captureStandardStreams(t, func() {
-			if err := ValidateConfiguration(conf, "pipeline view"); err != nil {
-				t.Fatalf("expected no error when only org is missing, got %v", err)
-			}
+			validationErr = ValidateConfiguration(conf, "pipeline view")
 		})
+
+		if validationErr != nil {
+			t.Fatalf("expected no error when only org is missing, got %v", validationErr)
+		}
 
 		if stdout != "" {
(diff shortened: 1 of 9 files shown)


pkg/cmd/validation/config_test.go
```diff
@@ -52,6 +55,26 @@ func TestValidateConfiguration_MissingValues(t *testing.T) {
...
+	t.Run("missing org warning is written to stderr", func(t *testing.T) {
+		t.Setenv("BUILDKITE_API_TOKEN", "token")
+		t.Setenv("BUILDKITE_ORGANIZATION_SLUG", "")
+		conf := newTestConfig(t)
+
+		stdout, stderr := captureStandardStreams(t, func() {
+			if err := ValidateConfiguration(conf, "pipeline view"); err != nil {
+				t.Fatalf("expected no error when only org is missing, got %v", err)
+			}
+		})
```
mcncl, line 67:
```suggestion
		var validationErr error
        stdout, stderr := captureStandardStreams(t, func() {
        	validationErr = ValidateConfiguration(conf, "pipeline view")
        })

		if validationErr != nil {
        	t.Fatalf("expected no error when only org is missing, got %v", validationErr)
        }
```

Kind of nitty, but if the output of the tests is redirected to a pipe then the output might be lost.

mcncl (approved):
Super minor thing which is more house-keepy than anything, but happy to approve/merge.

pkg/cmd/validation/config_test.go
mtibben: Addressed in 1dde33b

PR #792, opened 2026-04-17
Print build link in preflight summary

matthewborden:
### Description

Print a link to the build in the summary when preflight finishes.

### Testing
- [X] Tests have run locally (with `go test ./...`)
- [X] Code is formatted (with `go fmt ./...`)

### DIFF
--- a/cmd/preflight/render.go
+++ b/cmd/preflight/render.go
@@ -137,13 +137,18 @@ func summaryHeader(e Event) string {
 }
 
 func summaryBuildLine(e Event) string {
-	if e.BuildURL == "" {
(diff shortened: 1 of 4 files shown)


mcncl (approved):
This looks good. 

Minor note: including the full build URL in the TTY summary makes an existing narrow-terminal wrapping issue more likely to show up, since the final summary path doesn’t currently go through `hardwrapLine`.

Probably not a blocker for this PR, but worth  considering as follow-up.

catkins:
How does amp do those nice clickable links, or is it just handling the mouse events 🤔

PR #793, opened 2026-04-20
Preflight Test Results Summary

matthewborden:
This branch adds a server-aggregated test summary to bk preflight so the final output is more useful once a build completes.

The summary shows per-run results for the build, where each run maps to a test suite, and includes aggregated passed / failed / skipped counts. Failed tests are filtered to group-only-failed, so the summary excludes tests where at least one execution wasn't failed.

A new --await-test-results[=<duration>] flag lets bk preflight wait for Test Engine results to be processed before fetching the final summary. When enabled without a value, it waits for the default duration
before loading the aggregated results.

In interactive output, the TUI shows the top 5 failed test results alongside the per-run summary. In JSON output, the summary includes up to the top 100 failed test results.

This change depends on the server-side preflight summary implementation in buildkite/buildkite#29097 (https://github.com/buildkite/buildkite/pull/29097).


cmd/preflight/preflight.go
```diff
@@ -53,6 +56,31 @@ func preflightUserAgentSuffix() string {
 	return "buildkite-cli-preflight/" + major + ".x"
 }
 
+type awaitTestResultsFlag struct {
```
wolfeidau, line 59: Does this need to be exported, as in `AwaitTestResultsFlag`? or it works fine unexported?

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.25.0
 require (
 	github.com/alecthomas/kong v1.15.0
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be
-	github.com/buildkite/go-buildkite/v4 v4.21.0
+	github.com/buildkite/go-buildkite/v4 v4.20.0
```
wolfeidau: Downgrade?

wolfeidau (commented):
couple of low priority things.

go.mod
matthewborden: [a8845ce](https://github.com/buildkite/cli/pull/793/commits/a8845ceca3204fc4d55f6ec0cbde996232f8c410)

cmd/preflight/preflight.go
matthewborden: Not currently, the flag is only exposed on `bk preflight run` at the moment. We'll need to merge `bk preflight` and `bk preflight run`, but I'll do that in a seperate PR.

mcncl (approved):

PR #800, opened 2026-04-22
Tighten the Renovate config to be more defensive

mcncl:
### Description

Tightens the Renovate config and adds protection to CI against risky merges.

### Changes

- adds a `vulncheck` task to Mise
- implements a step in CI for `vulncheck` run
- adds `minimumReleaseAge` config to `renovate.json`


.buildkite/pipeline.yml
```diff
@@ -28,9 +28,17 @@ steps:
     plugins:
       - mise#v1.1.1: ~
 
+  - name: ":lock: govulncheck"
+    key: vulncheck
+    secrets:
+      - MISE_GITHUB_TOKEN
+    command: 'govulncheck ./...'
```
petetomasik: Shouldn't this be `mise run vulncheck` ?

renovate.json
```diff
@@ -9,6 +9,7 @@
   ],
   "labels": ["dependencies"],
   "postUpdateOptions": ["gomodTidy"],
+  "minimumReleaseAge": "3 days",
```
petetomasik, line 12: Can we bump this to 5-7 days instead?

mcncl: @petetomasik we can, but can I ask why? Is there any evidence to suggest 5 or 7 is better, from a CVE perspective? We could also bump it to 9, 11, etc, but at a certain point we're missing out on bug fixes.

petetomasik: @mcncl CVEs should be [configured](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts) to override `minimumReleaseAge` so that they are not lumped in with the proposed release age delay of dependencies:
```
  "vulnerabilityAlerts": {
    "enabled": true,
    "minimumReleaseAge": "0"
  }
```

.buildkite/pipeline.yml
```diff
@@ -28,9 +28,17 @@ steps:
     plugins:
       - mise#v1.1.2: ~
 
+  - name: ":lock: govulncheck"
+    key: vulncheck
+    secrets:
+      - MISE_GITHUB_TOKEN
+    command: 'mise run vulncheck'
+    plugins:
+      - mise#v1.1.1: ~
```
petetomasik: Plugin version bump:
```suggestion
      - mise#v1.1.2: ~
```

petetomasik (approved):

PR #801, opened 2026-04-22
automate the release notes grouping

mcncl:
### Description

Editing the release notes is annoying.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -11,6 +11,19 @@ release:
 
 changelog:
   use: github
+  sort: asc
+  abbrev: -1
+  groups:
+    - title: Added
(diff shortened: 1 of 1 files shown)


KSHITIZ6341 (commented):
Thanks for automating this; this should make releases easier.

I think the grouping may need attention before merge because ordinary user-facing `fix:` commits currently fall through to `Internal`. The only non-dependency user-facing group matches `feat`/`add`, while the dependency group handles deps, so commits like `fix: correct broken output` would be hidden under the catch-all internal group.

Suggested direction:
Add a `Fixed` or `Fixes and improvements` group for `fix(...)?:` before `Internal`, or make `Internal` explicitly match only non-user-facing prefixes such as `chore`, `refactor`, `test`, and `docs`.

PR #802, opened 2026-04-23
Preflight `--exit-on` & change default to exit on build failing state

matthewborden:
This introduces `--exit-on` handling for `bk preflight`, and changes watch mode to stop on the first observed `failing` build state by default instead of always waiting for a terminal build state.

The goal of preflight is to reduce time to actionable failure. Previously `bk preflight` would continue polling until the build reached a terminal state, even after Buildkite had already marked the build as `failing`. The failing state indicates that the build is on track to fail, all automatic retries have been exhausted and the step has hard failed.

The defaults for --exit-on is `--exit-on=build-failing` when `--watch` is enabled.

The previous behaviour of waiting for the final terminal state is available through `--exit-on=build-terminal` to preserve the previous watch-until-terminal behavior. `--exit-on` is not supported when `--watch=false`. Reuses the existing build cancel path so early exits cancel the remote build and cleanup the branch by default, but this can be skipped with `--no-cleanup`. Marks early-stop summaries as incomplete so output can distinguish “stopped on `failing`” from a completed terminal failure while still returning the existing incomplete-failure exit code.

### DIFF
--- a/cmd/preflight/preflight.go
+++ b/cmd/preflight/preflight.go
@@ -284,6 +284,17 @@ func (c *RunCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 		summaryEvent := newBuildSummaryEvent(preflightID.String(), pipelineName, build.Number, build.WebURL, finalBuild, startedAt)
 		summaryEvent.ApplySummaryMeta(summaryMeta{Incomplete: true, StopReason: "build-failing", BuildCanceled: buildCanceled})
 		summaryEvent.ApplyJobResults(finalBuild, tracker)
+		showResult, showErr := c.loadFinalResult(ctx, f.RestAPIClient, resolvedPipeline.Org, resolvedPipeline.Name, build.Number)
+		if showErr == nil {
+			summaryEvent.Tests = showResult.Tests
+		} else if globals.EnableDebug() {
+			_ = renderer.Render(Event{
+				Type:        EventOperation,
+				Time:        time.Now(),
+				PreflightID: preflightID.String(),
+				Title:       fmt.Sprintf("Debug: failed to load final test summary: %v", showErr),
+			})
+		}
 		_ = renderer.Render(summaryEvent)
 		cleanupBranch()
 		_ = renderer.Close()
--- a/cmd/preflight/preflight.go
+++ b/cmd/preflight/preflight.go
@@ -281,16 +281,10 @@ func (c *RunCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 			buildCanceled = cancelBuild(f, renderer, resolvedPipeline.Org, resolvedPipeline.Name, build.Number, preflightID.String(), globals.EnableDebug())
 		}
 
-		_ = renderer.Render(buildSummaryEvent(
-			finalBuild,
-			tracker,
-			preflightID.String(),
-			pipelineName,
-			build.Number,
-			build.WebURL,
-			startedAt,
-			summaryMeta{Incomplete: true, StopReason: "build-failing", BuildCanceled: buildCanceled},
-		))
+		summaryEvent := newBuildSummaryEvent(preflightID.String(), pipelineName, build.Number, build.WebURL, finalBuild, startedAt)
+		summaryEvent.ApplySummaryMeta(summaryMeta{Incomplete: true, StopReason: "build-failing", BuildCanceled: buildCanceled})
+		summaryEvent.ApplyJobResults(finalBuild, tracker)
+		_ = renderer.Render(summaryEvent)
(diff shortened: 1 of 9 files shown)


matthewborden:
@codex review

cmd/preflight/preflight.go
```diff
@@ -23,28 +23,37 @@ import (
...
-	Interval         float64              `help:"Polling interval in seconds when watching." default:"2"`
-	NoCleanup        bool                 `help:"Skip deleting the remote preflight branch after the build finishes."`
-	AwaitTestResults awaitTestResultsFlag `name:"await-test-results" help:"After the build finishes, wait for tests results to be processed by Test Engine. Provide a duration like 10s, or omit the value to wait 30s."`
-	Text             bool                 `help:"Use plain text output instead of interactive terminal UI." xor:"output"`
-	JSON             bool                 `help:"Emit one JSON object per event (JSONL)." xor:"output"`
+	Pipeline         string                         `help:"The pipeline to build. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}." short:"p"`
+	Watch            bool                           `help:"Watch the build until completion." default:"true" negatable:""`
+	ExitOn           []internalpreflight.ExitPolicy `help:"Exit when a condition is met. Options: build-failing (default, exits when a build enters the failing state), build-terminal (exits when the build reaches a terminal state)."`
```
mcncl, line 34: Using an [enum](https://github.com/alecthomas/kong#supported-tags) here for the options might assist humans and LLMs understand what is available as a flag value

internal/preflight/exit_policy.go
```diff
@@ -0,0 +1,47 @@
...
+	ExitOnBuildTerminal
+)
+
+func (p *ExitPolicy) UnmarshalText(text []byte) error {
+	value := string(text)
+	name, _, _ := strings.Cut(value, ":")
+	switch name {
+	case "build-failing":
+		*p = ExitOnBuildFailing
+	case "build-terminal":
```
mcncl, line 24: Note that `blocked` builds are classified as _incomplete_, which would mean that preflight could, AFAICT, wait indefinitely for a `blocked` build if a preflight build ends with a `block`

cmd/preflight/preflight.go
```diff
@@ -313,45 +334,75 @@ func (c *RunCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
...
+
+		_ = renderer.Render(Event{Type: EventOperation, Time: time.Now(), PreflightID: preflightID, Title: fmt.Sprintf("Warning: failed to cancel build #%d: %v", buildNumber, err)})
+		return false
+	}
+
+	_ = renderer.Render(Event{Type: EventOperation, Time: time.Now(), PreflightID: preflightID, Title: fmt.Sprintf("Cancelled build #%d", buildNumber)})
+	return true
+}
+
+func (c *RunCmd) loadFinalResult(ctx context.Context, client *buildkite.Client, org, pipeline string, buildNumber int) (internalpreflight.SummaryResult, error) {
```
mcncl, line 367: From the looks of it, `bk preflight --exit-on=build-failing --await-test-results` can still block for 30s+ after the build first enters failing because we always call `loadFinalResult()` in https://github.com/buildkite/cli/pull/802/changes#diff-84d84781dd3af35ddaac392e2c407af19c23cae3fb708d8e86e2eaf2b852e0c6R278

mcncl: Maybe consider an [xor](https://github.com/alecthomas/kong#supported-tags) on the flags?

cmd/preflight/preflight.go
matthewborden: I don't think an enum is appropriate because I'd like the CLI flag to support parameters eg. `tests-failed:2` in the reasonable soon future.

internal/preflight/exit_policy.go
matthewborden: I reckon we should address that separately of this PR, it's the existing behaviour.

cmd/preflight/preflight.go
matthewborden: That's an intentional decision, to try (with best effort) to show the test results of the step (if there is a TE run). Preflight exits when the build transitions to failed. The job that caused the build to fail may have submitted test results, but those may not have been processed by Test Engine by the time the job finishes, so we're waiting 30 seconds to allow those results to become available in Test Engine so Preflight can display them.

cmd/preflight/preflight.go
mcncl: So the goal is to have `--exit-on tests-failed:2`? Have we checked that's possible?

mcncl (approved):

PR #804, opened 2026-04-24
Disable in-progress test failures during preflight

matthewborden:
### Description

We're currently printing to the preflight log when a test fails, and then again when it subsequently passes. With the new test summary introduced in (https://github.com/buildkite/cli/pull/793), we can show tests which failed all executions at the end of a preflight run with a higher signal to noise ratio (eg. all retries have failed), and reduce the amount of context we omit to an LLM. 

I've only disabled test tracking for now, with some further iteration on excluding flaky tests we could bring back in-progress test failures.


mcncl (approved):
🫗

PR #810, opened 2026-04-29
feat: add `skills` command group for use with LLMs

mcncl:
### Description

Adds a command set for `skill` to enable adding Buildkite relevant skills to an LLM config

### Changes

- adds a `skill` command set with `add`, `update` and `delete` options

### Usage

```sh
# Install a skill locally (current dir)
bk skill add buildkite-cli

# Install globally; will look for ~/.cursor, ~/.claude, or both
bk skill add buildkite-cli --global

# Install to non-Claude/Cursor
bk skill add buildkite-cli --[path/to/location] ~/.pi/agents/skills
```

### Testing

### DIFF
--- a/cmd/skill/skill.go
+++ b/cmd/skill/skill.go
@@ -1,0 +1,423 @@
+package skill
+
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const (
+	defaultRepo   = "buildkite/skills"
+	defaultBranch = "main"
+)
+
+type AddCmd struct {
+	Name   string `arg:"" help:"Name of the Buildkite skill to install (for example, buildkite-api)."`
+	Agent  string `help:"Agent/harness to install for (claude or cursor). Auto-detected from .claude or .cursor by default." optional:""`
+	Global bool   `help:"Install globally in your home directory instead of the current project."`
+	Path   string `help:"Custom skills directory to install into, for agents such as Amp or Pi." type:"path" optional:"" aliases:"to,location"`
+	Force  bool   `help:"Overwrite an existing installed skill."`
+	Repo   string `help:"GitHub repository to install skills from." default:"${skill_repo}" hidden:""`
+	Branch string `help:"Git branch to install skills from." default:"${skill_branch}" hidden:""`
+}
+
+func (c *AddCmd) Help() string {
+	return `Install a Buildkite skill from github.com/buildkite/skills.
+
+By default, the target is auto-detected from a project .claude or .cursor
+folder. Use --agent to choose a target explicitly, --global to install to all
+existing global agent directories (~/.claude and/or ~/.cursor), or --path for
+another agent's skills directory.
+
+Examples:
+  # Install buildkite-api into the detected project agent
+  $ bk skill add buildkite-api
+
+  # Install for Claude Code in this project
+  $ bk skill add buildkite-api --agent claude
+
+  # Install globally for Cursor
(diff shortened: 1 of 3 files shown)


cmd/skill/skill.go
```diff
@@ -0,0 +1,387 @@
...
+		if err := os.Rename(extracted, dest); err != nil {
+			return err
+		}
+		fmt.Printf("Installed %s skill %q to %s\n", target.agent, name, dest)
+	}
+	return nil
+}
+
+func downloadRepoArchive(repo, branch, dest string) error {
+	url := fmt.Sprintf("https://codeload.github.com/%s/zip/refs/heads/%s", repo, branch)
```
catkins, line 345: TIL `codeload.github.com`

cmd/skill/skill.go
```diff
@@ -0,0 +1,387 @@
...
+`
+}
+
+func (c *DeleteCmd) Run() error {
+	targets, err := resolveTargets(c.Agent, c.Global, c.Path)
+	if err != nil {
+		return err
+	}
+	for _, target := range targets {
+		dest := filepath.Join(target.SkillsDir(), c.Name)
```
matthewborden, line 171: This should perform validation that the skill name is a literal string, similar to update: https://github.com/buildkite/cli/pull/810/changes#diff-aa6f22a0a6f8c5ff110b115bbba9296f540484bc4408ba11f6aba2895d232f15R263-R265.

cmd/skill/skill.go
```diff
@@ -0,0 +1,387 @@
...
+		if err := os.RemoveAll(dest); err != nil {
```
matthewborden, line 172: Should we check if the target path exists first before trying to delete it, or just persist?

cmd/skill/skill.go
```diff
@@ -0,0 +1,387 @@
...
+				return err
+			}
+		}
+		if len(installedTargets) == 0 {
+			return fmt.Errorf("skill %q is not installed for any selected target", c.Name)
+		}
+		return installSkillToTargets(c.Name, installedTargets, true, c.Repo, c.Branch)
+	}
+
+	seen := map[string]bool{}
```
matthewborden:
```suggestion
	seenSkillNames := map[string]bool{}
```

cmd/skill/skill.go
```diff
@@ -0,0 +1,387 @@
...
+			if info, err := os.Stat(filepath.Join(target.SkillsDir(), name)); err == nil && info.IsDir() {
+				installedTargets = append(installedTargets, target)
+			} else if err != nil && !os.IsNotExist(err) {
+				return err
+			}
+		}
+		if err := installSkillToTargets(name, installedTargets, true, c.Repo, c.Branch); err != nil {
+			return err
+		}
+	}
```
matthewborden: I think this might download the skills repo N times for N installed skills. If there's an entry for each skill in `seen`, then this calls `installSkillToTargets` which calls `downloadRepoArchive` for each skill. Could we invert this and download the archive once and pass it into `installSkillToTargets`?

cmd/skill/skill.go
```diff
@@ -0,0 +1,423 @@
...
+		rc.Close()
+		if copyErr != nil {
+			return copyErr
+		}
+		if closeErr != nil {
+			return closeErr
+		}
+	}
+	if !found {
+		return fmt.Errorf("skill %q not found in github.com/%s", skillName, defaultRepo)
```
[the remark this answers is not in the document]
mcncl: We only want to use the `buildkite/skills` repo

matthewborden (approved):

PR #820, opened 2026-05-04
feat: allow for preselecting org when authing

mcncl:
### Description

We recently added support for passing in the `organization` or
`organization_uuid` as a URL parameter, which means that we can
preselect the organization in the OAuth modal on behalf of the user.
This "fixes" an issue where the most recently _interacted with_
organization would be selected.

### Changes

- allows users to set `--org` _without_ providing `--token`, too
- passed the set `--org` value as a URL parameter to the Buildkite OAuth modal
- overrides current behaviour of _select most recently interacted with_

### Testing

### DIFF
--- a/cmd/auth/login.go
+++ b/cmd/auth/login.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/alecthomas/kong"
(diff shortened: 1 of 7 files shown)


cmd/auth/login.go
```diff
@@ -18,8 +19,18 @@ import (
...
+}
+
+var uuidPattern = regexp.MustCompile(`(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
+
+func organizationIdentifier(org string) (orgSlug, orgUUID string) {
+	if uuidPattern.MatchString(org) {
+		return "", org
+	}
+
+	return org, ""
```
scadu: I'd rather use existing parsers, e.g. https://github.com/google/uuid.

```suggestion
func organizationIdentifier(org string) (orgSlug, orgUUID string) {
      if _, err := uuid.Parse(org); err == nil {
          return "", org
      }
      return org, ""
  }
```

scadu (approved):

PR #828, opened 2026-05-12
feat: Add device auth login

lox:
## Problem

`bk auth login` currently assumes a browser can be opened on the same machine that is running the CLI and that a localhost callback can be reached. That is awkward or impossible in common headless cases, especially SSH sessions, remote development boxes, and other environments where the user can authorize in a separate browser but cannot rely on loopback browser OAuth.

Buildkite now has OAuth device authorization endpoints, so the CLI should be able to start a device flow and poll for the resulting access token without opening a browser callback listener.

## Impact

Users can run `bk auth login --device` on a headless or remote machine, authorize from another browser, and have the CLI persist the resulting access and refresh tokens through the same OAuth storage path as browser login. Existing browser login and `--token` login behavior is preserved.

## What Changed

- Added `bk auth login --device`.
- Added OAuth device authorization request and device-code token polling helpers.
- Reused the existing OAuth persistence path so browser login and device login both resolve the authorized organization, store the access token, store refresh tokens when present, and select the organization in config.
- Only advertises automatic refresh when refresh-token storage actually succeeds.
- Rejects incompatible `--device --token` and `--device --org` combinations; device flow chooses the organization on the authorization page.
- Added request-level tests for device authorization, `invalid_scope`, `authorization_pending`, and `slow_down` polling behavior.
- Added a command-level device-login test that exercises device authorization, token exchange, organization lookup, keyring storage, refresh-token storage, and config selection.
- Updated the pinned Go toolchain to 1.26.4 and removed unused `go-buildkite/v5` module metadata inherited from the rebase so the current tidy and `govulncheck` CI gates pass.

### DIFF
--- a/pkg/oauth/oauth.go
+++ b/pkg/oauth/oauth.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -19,6 +20,10 @@ import (
 
 const (
 	DefaultHost = "buildkite.com"
+
+	deviceCodeGrantType     = "urn:ietf:params:oauth:grant-type:device_code"
+	defaultDeviceInterval   = 5 * time.Second
+	deviceSlowDownIncrement = 5 * time.Second
 )
 
 // AllScopes is the complete set of Buildkite API token scopes. When no --scopes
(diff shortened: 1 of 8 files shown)


pkg/oauth/oauth.go
```diff
@@ -370,6 +374,160 @@ func (f *Flow) ExchangeCode(ctx context.Context, code string) (*TokenResponse, e
...
+	for {
+		tokenResp, err := exchangeDeviceCode(ctx, cfg, deviceAuth.DeviceCode)
+		if err == nil {
+			return tokenResp, nil
+		}
+
+		var tokenErr *tokenError
+		if !errors.As(err, &tokenErr) {
+			return nil, err
+		}
```
JoeColeman95, line 451: Not blocking, but since device login here is built for headless/remote/SSH, I think handling transient failures as non-fatal is the better way of handling this rather than aborting the whole flow. A dropped request can force a new flow entirely currently.

I'd say keeping context cancellation/deadline as-is is for sure the play, though. Everything else, I think it makes sense to wait out the interval and retry:


```suggestion
		var tokenErr *tokenError
		if !errors.As(err, &tokenErr) {
			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
				return nil, err
			}
			if err := sleep(ctx, interval); err != nil {
				return nil, err
			}
			continue
		}
```

JoeColeman95 (approved):

pkg/oauth/oauth.go
lox: Fixed in 5662469: device-code polling now retries transient transport errors after the current interval while still returning immediately on context cancellation/deadline. Added tests for both paths.

PR #832, opened 2026-05-19
fix(http): skip AuthTransport bearer on redirected hops

### LINKED ISSUE #830: [Bug]: bk artifact download unconditionally adds a token to pre-signed URL downloads (breaking downloads)
### What happened?

AuthTransport.RoundTrip (internal/http/refresh_transport.go:54) unconditionally sets Authorization: Bearer <token>. When the API 302-redirects an artifact download to a presigned buildkiteartifacts.com URL, Go's stdlib correctly strips the inherited bearer — but AuthTransport runs again on the redirected hop and re-injects it. S3 sees both X-Amz-Signature=… and Authorization and returns:

> 400 InvalidArgument: Only one auth mechanism allowed

Also it leaks your BK token.

### Version

`bk version 3.42.0`

### What environment are you seeing the problem on?

macOS

SAY-5:
`AuthTransport.RoundTrip` unconditionally re-set the Authorization header, which clobbered stdlib `http.Client`'s cross-host strip on artifact-download redirects to presigned `buildkiteartifacts.com` / S3 URLs. The presigned URL already carries its own auth via `X-Amz-Signature`, so the extra header either leaks the bearer or breaks the download with S3's "Only one auth mechanism allowed" 400. New test redirects through two `httptest` servers and asserts the second hop sees no `Authorization`.

Closes #830


internal/http/refresh_transport.go
```diff
@@ -52,9 +52,12 @@ type AuthTransport struct {
 }
 
 func (t *AuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	token := t.TokenSource.Token()
-	if token != "" {
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+	// Skip re-injecting on redirected hops so stdlib's cross-host strip stays in effect.
```
mcncl, line 55: @SAY-5 thanks for the PR! As far as I can tell, this only fixes the normal redirect hop, which is fine (and a nice fix). I'm wondering if you wanted to additionally include a fix for `401` responses, which will still use the `RefreshTransport`

mcncl: CI is failing due to an expired token, I think, so I'll look in to fixing that up

mcncl (approved):

PR #833, opened 2026-05-20
fix: Use organization job routes

lox:
### Description

`bk job log`, `bk job reprioritize`, and `bk job unblock` all take a job UUID, but the REST-backed command paths still required pipeline or build context. That made `bk job log <job-uuid>` fail with `failed to resolve a pipeline` outside a pipeline checkout, and forcing callers to pass pipeline/build identifiers would change the existing UUID-only command shape.

The CLI now calls the new organization-level job REST routes directly under the selected organization. This keeps UUID-only job commands working, avoids scanning builds, and uses the same API surface we added in buildkite/buildkite#29824. The existing `--pipeline` and `--build` flags on `job log` and `job reprioritize` remain accepted for compatibility, but now emit a stderr warning when supplied because they are ignored.

### Changes

- Add shared REST helpers for `/v2/organizations/:org/jobs/:job_id/{log,reprioritize,unblock}`.
- Update `bk job log` to fetch logs directly by job UUID.
- Update `bk job reprioritize` to reprioritize directly by job UUID.
- Update `bk job unblock` to unblock directly by job UUID while preserving `--data` and stdin fields.
- Require a selected organization before constructing organization-scoped job route requests.
- Warn when deprecated `--pipeline` or `--build` context flags are supplied to UUID-only job commands.
- Update command tests to assert the organization-level paths, methods, headers, request bodies, missing-organization guard, deprecated flag warning, and already-unblocked REST error handling.

Relevant examples now work without pipeline or build context:

```text
bk job log 019e6209-ac2a-4d32-84c7-2abcee10099c
bk job reprioritize 019e6209-ac2a-4d32-84c7-2abcee10099c 10
bk job unblock 019e6209-ac2a-4d32-84c7-2abcee10099c --data '{"name":"value"}'
```

### Testing

Additional checks run:

- `mise exec -- go test ./cmd/job`
- `mise exec -- go test ./cmd/job ./internal/pipeline/resolver ./internal/build/resolver`
- `mise run build`
- `mise run test`
- `mise run lint`
- live `./dist/bk job log 019e6209-ac2a-4d32-84c7-2abcee10099c --no-pager`
- live safe `./dist/bk job unblock 019e6209-ac2a-4d32-84c7-2abcee10099c --no-pager` returned `400 This job type cannot be unblocked`
- live safe `./dist/bk job reprioritize 019e6209-ac2a-4d32-84c7-000000000000 10 --no-pager` returned `404 No job found`

### Disclosures / Credits

Codex implemented and validated this update with human direction. The organization-level REST API routes were added separately in buildkite/buildkite#29824.

### DIFF
--- a/cmd/job/log.go
+++ b/cmd/job/log.go
@@ -6,33 +6,27 @@ import (
 	"regexp"
 
 	"github.com/alecthomas/kong"
-	buildResolver "github.com/buildkite/cli/v3/internal/build/resolver"
-	"github.com/buildkite/cli/v3/internal/build/resolver/options"
 	"github.com/buildkite/cli/v3/internal/cli"
 	bkIO "github.com/buildkite/cli/v3/internal/io"
-	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
 	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 	"github.com/buildkite/cli/v3/pkg/cmd/validation"
 )
 
 type LogCmd struct {
 	JobID        string `arg:"" help:"Job UUID to get logs for"`
-	Pipeline     string `help:"The pipeline to use. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}" short:"p"`
-	BuildNumber  string `help:"The build number" short:"b"`
+	Pipeline     string `help:"Deprecated; ignored because job UUIDs are scoped to the selected organization" short:"p"`
+	BuildNumber  string `help:"Deprecated; ignored because job UUIDs are scoped to the selected organization" short:"b"`
 	NoTimestamps bool   `help:"Strip timestamp prefixes from log output" name:"no-timestamps"`
 }
 
 func (c *LogCmd) Help() string {
 	return `
 Examples:
-  # Get a job's logs by UUID (requires --pipeline and --build)
-  $ bk job log 0190046e-e199-453b-a302-a21a4d649d31 -p my-pipeline -b 123
-
-  # If inside a git repository with a configured pipeline
-  $ bk job log 0190046e-e199-453b-a302-a21a4d649d31 -b 123
+  # Get a job's logs by UUID
+  $ bk job log 0190046e-e199-453b-a302-a21a4d649d31
 
   # Strip timestamp prefixes from output
-  $ bk job log 0190046e-e199-453b-a302-a21a4d649d31 -p my-pipeline -b 123 --no-timestamps
+  $ bk job log 0190046e-e199-453b-a302-a21a4d649d31 --no-timestamps
 `
 }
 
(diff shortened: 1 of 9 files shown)


Megh03 (approved):
LGTM @lox ! Was just a little curious on the pagination side, like for large orgs with many builds, with the current code . the job lookup could paginate through a lot of pages before finding the job . But I think it's a fair tradeoff to make 🤔

lox:
Yeah, I don't love the pagination approach, I don't think it will be viable. 

The challenge is that currently this command doesn't need a pipeline identifier, as the GraphQL endpoint is job UUID only, where as the REST API needs `/v2/organizations/:org/pipelines/:pipeline/builds/:build/jobs/:job_id/unblock`.

I think realistically we either need to require an optional `--pipeline` and switch between GraphQL and REST (which would be gross), or add a new REST endpoint for this.

lox:
Converting back to draft whilst we figure it out.

lox:
@buildsworth-bk review please

cmd/job/reprioritize.go
```diff
@@ -18,18 +15,15 @@ import (
 type ReprioritizeCmd struct {
 	JobID       string `arg:"" help:"Job UUID to reprioritize"`
 	Priority    int    `arg:"" help:"New priority value for the job"`
-	Pipeline    string `help:"The pipeline to use. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}" short:"p"`
-	BuildNumber string `help:"The build number" short:"b"`
+	Pipeline    string `help:"Deprecated; ignored because job UUIDs are scoped to the selected organization" short:"p"`
```
scadu: Not a blocker, same as https://github.com/buildkite/cli/pull/833/changes#r3301441665. Perhaps worth emitting a warning here?

cmd/job/rest.go
scadu, line 1: Not a blocker, but I wonder if we plan to consolidating this helper(?) into go-buildkite?

cmd/job/unblock.go
```diff
@@ -58,59 +54,68 @@ func (c *UnblockCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
...
+	if fields == nil {
+		return nil, fmt.Errorf("unblock data must be a JSON object")
+	}
+
+	return fields, nil
+}
+
+func isAlreadyUnblocked(err error) bool {
+	var apiErr *buildkite.ErrorResponse
+	return errors.As(err, &apiErr) && apiErr.Message == "The job's state must be blocked"
```
scadu: Looks like backend also provides `The job's state must be blocked` which should be also handled.

cmd/job/log.go
```diff
@@ -6,33 +6,27 @@ import (
...
 	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 	"github.com/buildkite/cli/v3/pkg/cmd/validation"
 )
 
 type LogCmd struct {
 	JobID        string `arg:"" help:"Job UUID to get logs for"`
-	Pipeline     string `help:"The pipeline to use. This can be a {pipeline slug} or in the format {org slug}/{pipeline slug}" short:"p"`
-	BuildNumber  string `help:"The build number" short:"b"`
+	Pipeline     string `help:"Deprecated; ignored because job UUIDs are scoped to the selected organization" short:"p"`
+	BuildNumber  string `help:"Deprecated; ignored because job UUIDs are scoped to the selected organization" short:"b"`
```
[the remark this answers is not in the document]
lox: Fixed in 97a9925: `bk job log` now writes a one-line stderr warning when either deprecated context flag is set, while still accepting the flags for compatibility.

cmd/job/reprioritize.go
lox: Fixed in 97a9925: `bk job reprioritize` uses the same deprecated-context warning path as `job log`.

cmd/job/log.go
```diff
@@ -53,40 +47,12 @@ func (c *LogCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
...
 	var logContent string
 	if err = bkIO.SpinWhile(f, "Fetching job log", func() error {
-		jobLog, _, apiErr := f.RestAPIClient.Jobs.GetJobLog(
+		jobLog, apiErr := getJobLog(
 			ctx,
-			bld.Organization,
-			bld.Pipeline,
-			fmt.Sprint(bld.BuildNumber),
+			f.RestAPIClient,
+			f.Config.OrganizationSlug(),
```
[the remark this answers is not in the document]
lox: Fixed in 97a9925: the org-scoped job commands now require a selected organization before building the REST path, so missing org fails locally with a clear error instead of calling `/v2/organizations//jobs/...`.

cmd/job/unblock.go
lox: Fixed in 97a9925: the already-unblocked handling now explicitly covers `The job's state must be blocked` from REST, including the trailing-period variant, with test coverage.

cmd/job/rest.go
lox: Agreed. I kept these helpers local and small in this PR because go-buildkite does not expose the new org-level job routes yet. Once it does, we can swap these for SDK methods without changing the command behavior.

lox:
@codex review

lox:
Review follow-up: I checked the risk labels and this repo currently only has `risk:l1`, so I left the label unchanged. The inline feedback is addressed in 97a9925.

scadu (approved):

PR #854, opened 2026-06-05
feat(config): add credential_store key

ivannalisetska:
- Add credential_store config key (auto, keyring, shm) with precedence env > local > user
- Propagate the configured value into BUILDKITE_CREDENTIAL_STORE at startup so bk auth login uses it without --credential-store
- Update help text on bk config set/get and bk auth login

TASK: SUP-7238

### Changes

Add credential_store key to bk config (auto, keyring, shm).
Wire it through bk config set/get/list/unset with validation.
Export the configured value into BUILDKITE_CREDENTIAL_STORE at process start so every command (notably bk auth login) picks it up without flags.
Precedence: --credential-store flag > env > local config > user config > auto.
Update help text on bk config set/get and bk auth login.
Unit tests for the new key, precedence chain, and the env bridge.

Can skip if changes are simple or clear from the commit messages.
-->

### Testing
- [x] Tests have run locally (with `go test ./...`) 39 packages, all ok, zero failures.
- [x] Code is formatted (with `go fmt ./...`) no files needed reformatting.


main.go
```diff
@@ -302,6 +303,10 @@ func run() int {
 
 	conf := config.New(nil, nil)
 
+	// Must run before kong dispatches: bk auth login builds its keyring
+	// before factory.New() does.
+	factory.ApplyCredentialStoreFromConfig(conf)
```
mcncl, line 308: Hmm, I don't think this would recognise a local config unless `--credential-store` is also passed, eg `bk auth login --local` wouldn't respect a local `.bk.yaml` config being present.

We'd need to ensure `login` respects `--local` config, or only support user configs.

ivannalisetska: thanks Ben! makes sense, credential_store is now user-only, matching the existing pattern for no_input, pager, telemetry, and experiments. Pushed as a follow-up commit

mcncl (approved):
Looks good! The changes from `main` will need merging in so that we can ensure the build is successful for the binaries.

PR #872, opened 2026-06-16
chore(deps): use our internal renovate config as base

mcncl:
### Description

Opts to using the internal Renovate config and so some existing config can be removed.


wolfeidau (approved):
👍🏻

PR #880, opened 2026-06-18
fix(http): query param normalisation in branch

mcncl:
### Description

Fixes the use of params in `browse` command so that they get normalised in full as the branch name

### DIFF
--- a/cmd/browse/browse.go
+++ b/cmd/browse/browse.go
@@ -3,6 +3,7 @@ package browse
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


cmd/browse/browse.go
```diff
@@ -204,5 +205,5 @@ func settingsURL(org, pipeline string) string {
 }
 
 func pipelineBranchURL(org, pipeline, branch string) string {
-	return fmt.Sprintf("https://buildkite.com/%s/%s?branch=%s", org, pipeline, branch)
+	return fmt.Sprintf("https://buildkite.com/%s/%s?branch=%s", org, pipeline, url.QueryEscape(branch))
```
pda, line 208: 🤖 suggestion was:

> build the query with `net/url`, e.g. `url.Values{"branch": []string{branch}}.Encode()`, or at minimum `url.QueryEscape(branch)`

I think the latter is fine 👍🏼

pda (approved):

PR #885, opened 2026-06-21
Adds an `--agent` flag to `job log` for the robots

mcncl:
### Description

Fetching a Buildkite job log for an LLM/agent to read is wasteful: the raw log is full of ANSI colour codes, inline `\x1b_bk;t=...` timestamps, and carriage-return progress redraws that burn tokens without adding meaning.

This PR adds an opt-in `--agent` flag to `bk job log` that digests the log into a compact, LLM-friendly form via the [`terminal-to-llm`](https://github.com/mcncl/terminal-to-llm) library (strips escapes/timestamps, resolves CR redraws, collapses duplicate/progress lines, and applies failure-focused windowing). On a real job log this cut output by ~87%.

To see the reduction on any job:

```sh
go build -o /tmp/bk .
echo "raw:   $(/tmp/bk job log <uuid> | wc -c) bytes"
echo "agent: $(/tmp/bk job log <uuid> --agent | wc -c) bytes"
```

### Changes

- Add `--agent` to `bk job log`, routing output through digest.Process
- Add supporting flags (only meaningful with `--agent`):
    - `--format plain|markdown` - render groups as Markdown headings/code blocks.
    - `--max-tokens N` — hard ceiling on estimated tokens (0 = unlimited).
    - `--no-window` — disable failure-focused windowing (keep all lines).

`bk job log --help`:

```
--no-timestamps     Strip timestamp prefixes from log output
--agent             Format output to be optimal for LLM consumption
                    (strips ANSI, deduplicates loops)
--format="plain"    Output rendering for `--agent`: plain or markdown
--max-tokens=INT    Hard ceiling on the estimated token count of `--agent` output (0 = unlimited)
--no-window         Disable failure-focused windowing in `--agent` output
```

### Testing

- [x] Tests have run locally (with `mise test`)
- [x] Code is formatted (with `mise format`)

Also verified end-to-end against a real job log across each flag combination.


cmd/job/log.go
```diff
@@ -17,6 +18,7 @@ type LogCmd struct {
 	Pipeline     string `help:"Deprecated; ignored because job UUIDs no longer require pipeline or build context" short:"p"`
 	BuildNumber  string `help:"Deprecated; ignored because job UUIDs no longer require pipeline or build context" short:"b"`
 	NoTimestamps bool   `help:"Strip timestamp prefixes from log output" name:"no-timestamps"`
+	LLMOptimized bool   `help:"Format output to be optimal for LLM consumption (strips ANSI, deduplicates loops)" name:"agent"`
```
mitchbne: Slightly bike-shedding here, but `agent` feels like it might be too generic (especially given 'agent' in Buildkite often refers to a process running the `buildkite-agent` CLI). What are your thoughts on naming this 'llm' instead?


```suggestion
	LLMOptimized bool   `help:"Format output to be optimal for LLM consumption (strips ANSI, deduplicates loops)" name:"llm"`
```

mcncl: @mitchbne I'm neither here nor there on it; other tools such as `coderabbit` use `--agent`, so I was initially presenting the user with an option that would be familiar. 

The flag would _likely_ only be used by an LLM, too, so it _may_ not be an issue for users; the LLM would "see" that there is an `--agent` command for _it_, likely after running `--help`.

But, I'm not about to die on the hill for that and I'm fine either way.

mcncl: I think `aliases: "llm"` should be fine here

mitchbne: Sounds good to me ✅

omehegan (approved):
Looks good, I previously reviewed the `terminal-to-llm` code so I get how that's integrating here. Test coverage seems fine, and looking at the way `terminal-to-llm` is used, I don't see any possible unhandled issues. Let's go!

PR #900, opened 2026-07-07
fix(auth): persist auth outside of repo

mcncl:
### Description

Auth currently breaks if you move outside of a project or delete/move `.bk.yaml`.

### Changes

- persists `auth` using the `~/.config/bk.yaml` instead of local, so moving to a non-BK project doesn't break login

### Testing


wolfeidau (approved):
👍🏻

PR #910, opened 2026-07-14
feat(job): add exclusion options

mcncl:
### Description

The Buildkite REST API supports `exclude_jobs` and `exclude_pipeline` query params to skip serializing the (potentially huge) jobs array and pipeline object in build responses. The [MCP server already uses this](https://github.com/buildkite/buildkite-mcp-server/blob/fdda11fad2ac7b9eb78b67a7dfb5af8b1efc53cd/pkg/buildkite/builds.go#L160-L166), and support exists in go-buildkite v5 (`BuildsListOptions.ExcludeJobs` / `ExcludePipeline`).

This PR applies these options everywhere the CLI fetches builds but never reads the jobs and/or pipeline payloads. No user-facing behaviour changes, just smaller, faster API responses, especially for builds with many jobs and for polling loops (`bk build watch`, `bk preflight`).

### Changes

Exclude **jobs + pipeline** (neither is read):
- `internal/build/resolver/with_options.go` - build resolution only uses the build number
- `internal/preflight/branch_build.go` - `ResolveBuilds` only uses build state/branch
- `cmd/preflight/preflight.go` (`loadFinalResult`) - only uses build ID + Test Engine data
- `cmd/build/list.go` - text output only; `-o json/yaml` keeps full payloads for compatibility

Exclude **pipeline** only (jobs still needed):
- `cmd/job/list.go` - jobs are extracted from builds; pipeline object unused
- `cmd/build/download.go` - needs jobs for log downloads
- `internal/build/watch/watch.go` - polling never reads the pipeline payload (now applied on every poll)

`bk build view` is intentionally untouched (renders jobs, and JSON output should stay complete)

### Testing

### DIFF
--- a/cmd/build/download.go
+++ b/cmd/build/download.go
@@ -16,6 +16,7 @@ import (
 	pipelineResolver "github.com/buildkite/cli/v3/internal/pipeline/resolver"
 	"github.com/buildkite/cli/v3/pkg/cmd/factory"
 	"github.com/buildkite/cli/v3/pkg/cmd/validation"
+	buildkite "github.com/buildkite/go-buildkite/v5"
 )
 
 type DownloadCmd struct {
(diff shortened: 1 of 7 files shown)


SorchaAbel (approved):
Nice one!

PR #927, opened 2026-07-28
Fix job unblock data precedence

### LINKED ISSUE #926: job unblock --data is ignored when stdin is a non-TTY with no data
### Description

`bk job unblock --data ...` ignores the explicit `--data` value when stdin is a non-TTY, even when stdin contains no data. The resulting unblock request is sent without fields (`{}`).

This matters for CI runners and other automation where stdin is commonly connected to an empty pipe or `/dev/null` rather than a TTY.

### Reproduction

Given a blocked job with fields:

```sh
bk job unblock "$JOB_ID" \
  --data '{"field-name":"value"}' \
  --yes --no-input --debug </dev/null
```

The debug output shows that the request body is `{}`, and the API reports missing/invalid fields. Running the same command in an interactive TTY uses `--data` correctly.

### Expected behavior

An explicitly supplied `--data` value should take precedence over stdin and be sent as the unblock fields, regardless of whether stdin is a TTY.

### Cause

`UnblockCmd.unblockFields` checks `HasDataAvailable(os.Stdin)` before checking `c.Data`. For an `*os.File`, `HasDataAvailable` returns true whenever the descriptor is not a terminal; it does not establish that bytes are actually available. With empty non-TTY stdin, the command reads an empty string, returns no fields, and never examines `--data`.

The relevant code appears unchanged in v3.54.0, v3.54.1, and current `main`.

### Suggested fix

Prefer explicit `--data` when it is non-empty, and only inspect/read stdin when `--data` was not supplied. It may also be useful to add a test with stdin connected to an empty non-TTY reader.

### Workarounds

Piping the JSON through stdin works:

```sh
printf '%s\n' '{"field-name":"value"}' |
  bk job unblock "$JOB_ID" --yes --no-input
```

Calling the REST endpoint through `bk api` also works.

nobradov1c:
## Summary

- prefer an explicitly supplied `--data` value over stdin in `bk job unblock`
- preserve piped stdin as the fallback when `--data` is absent
- add regression coverage for `--data` precedence, stdin fallback, and no-input behavior

Fixes #926.

## Testing

- `go test -race ./cmd/job`
- `go vet ./cmd/job`
- `git diff --check`

I also ran `go test ./...`; all relevant packages passed, but the existing keyring tests failed because this development environment has an active shared-memory credential store.


cmd/job/unblock.go
```diff
@@ -89,14 +89,16 @@ func (c *UnblockCmd) Run(kongCtx *kong.Context, globals cli.GlobalFlags) error {
 }
 
 func (c *UnblockCmd) unblockFields() (map[string]any, error) {
```
Damilola-obasa: Nice work!
One small suggestion on the test though, right now it works by temporarily swapping out the program's input with a placeholder one, then switching it back after the test is done. That works, but I think it would  be safer to just enter the test input directly into the function instead of swapping anything out. The reason is that if another test happens to run around the same time, they could end up interfering with each other.

cmd/job/unblock_test.go
```diff
@@ -86,6 +87,28 @@ func TestParseUnblockFields(t *testing.T) {
 	}
 }
 
+func TestUnblockFieldsPrefersDataOverEmptyNonTTYStdin(t *testing.T) {
```
Damilola-obasa, line 134: The other thing is, right now there's only one test, and it covers the case where both `--data` and manually entered data are given, confirming that `--data` correctly wins. But this change touches two other situations as well, when data is entered manually with no `--data` at all, and when neither one is provided, so it'd be worth adding tests for those too. That way we can be confident the reordering didn't affect anything that already worked before this fix. 
None of this is blocking, just some points to improve on before it merges

cmd/job/unblock.go
nobradov1c: Done.

cmd/job/unblock_test.go
nobradov1c: Updated.

nobradov1c:
Cleaned up the PR, sorry for the delay.

cmd/job/unblock_test.go
```diff
@@ -86,6 +88,48 @@ func TestParseUnblockFields(t *testing.T) {
...
+			want: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := UnblockCmd{Data: tt.data}
+			got, err := cmd.unblockFields(strings.NewReader(tt.stdin))
```
Damilola-obasa, line 123: One small non-blocking thought, it looks like the case using a real file (like /dev/null) for stdin got dropped along the way, and that was actually the setup that reproduced the original bug. Right now everything runs through sample text, which works fine for most cases, but in a real automated environment, it may not behave the same way.

nobradov1c: Added a test with os.DevNull

Damilola-obasa:
Thanks a lot! I believe this should be all good now. The status check would likely be resolved if you update the branch by merging main to this branch.

nobradov1c:
> Thanks a lot! I believe this should be all good now. The status check would likely be resolved if you update the branch by merging main to this branch.

Updated, thanks!

Damilola-obasa (approved):

PR #935, opened 2026-08-03
Prevent PostHog telemetry from delaying commands

SorchaAbel:
### Description

PostHog's Go client can wait for its 10-second upload timeout when the telemetry endpoint is unreachable. Because `bk` flushes telemetry during process shutdown, this can add approximately 10 seconds to normal CLI commands; disabling telemetry reduced an affected command to approximately 0.1 seconds.

Configure a 500 ms PostHog shutdown deadline so telemetry remains best-effort without materially delaying command exit. Healthy uploads still complete normally, while stalled uploads are cancelled when the deadline expires.

Context: [PB-2556](https://linear.app/buildkite/issue/PB-2556)

### Changes

- Bound PostHog client shutdown to 500 ms.
- Add regression coverage using an upload that starts but never responds.
- Verify the regression test fails at the previous 10-second timeout and completes in approximately 500 ms with the fix.

### Testing

- [x] Code is formatted (with `gofumpt`)
- [x] Focused regression test passes 10 times under the race detector
- [x] `golangci-lint` reports no issues for `pkg/analytics`

### Rollback

Revert this pull request to restore the PostHog SDK's default shutdown behavior. There are no migrations or persisted-data changes.

### Disclosures / Credits

Amp investigated the telemetry lifecycle, implemented the timeout and regression test, ran validation, and performed an iterative independent review. Sorcha Abel provided the real-world reproduction and product context.

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -5,9 +5,12 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/posthog/posthog-go"
 )
+
+const shutdownTimeout = 500 * time.Millisecond
 
 var (
 	// Set via -ldflags at build time: -X github.com/buildkite/cli/v3/pkg/analytics.apiKey=...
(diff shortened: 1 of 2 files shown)


mcncl:
Already fixed in https://github.com/buildkite/cli/pull/931

PR #940, opened 2026-08-07
Update job list to server-side filtering where possible

JoeColeman95:
### Description

`bk job list --queue test-queue --state running` takes forever to return, or never. The queue filter already ran on the server but the state filter didn't, so pages came back unfiltered and running was applied locally while the pager kept asking for more until it had 100 matches. A queue with fewer running jobs than that walks its whole history and never stops.

`organization.jobs` takes `state` and `clusterQueue` together, so this just sends the state with the queue.

I left `--since` and `--until` alone because the server does those as `DATE(jobs.created_at)`, which would be less accurate than what we do now.

`--duration` has no server side equivalent so is staying as a client side filter, so the queue pager now stops after 2000 jobs while a local filter is still throwing results away.

### Changes

- Refreshed schema as it was drastically outdated
- Only wire change is the two `$state` additions
 - No longer collapses `ASSIGNED` and `ACCEPTED` into `scheduled`, or `CANCELING` into `canceled`. That changes displayed output and it's needed, otherwise `--state` assigns returns rows and then filters them out.

### Testing

### Disclosures / Credits

Tests were written by Amp, I previously traced this on a ticket from a customer and established the fix then.


mitchbne:
Hello I don't have much context on this area of the code base. Would you be able to help me understand why we are doing these filters client side rather than server side?

JoeColeman95:
Hey @mitchbne,

I wrote this a while back following a customer request and our REST and GraphQL API's never supported filtering by state or queue then. We're talking ~1 year ago now. I since noticed that we'd upgraded the edges available, so have opted to upgrade this here to provide this functionality server-side opposed to handling it client-side as it can be rather slow for customers that have a lot of builds/jobs.

The change here is off the back of upgrades to bk/bk following a PF push, which began from inefficiencies for cluster queue insights IIRC?

But tl;dr, we had this requested a while back, it wasn't supported via the API, now it is.

tomowatt (approved):
I saw 14K additions and 5k deletions and wondered was the entire CLI being re-written but then saw it was the GraphQL Schema and `generated.go`

cmd/job/list.go
```diff
@@ -49,14 +51,13 @@ func (c *ListCmd) Help() string {
...
 can be passed with --pipeline or resolved from the current repository or config.
 
-Client-side filters: --queue, --state, --duration
-Server-side filters: --pipeline, --build, --step-key, --group-key, --since, --until
+Server-side filters: --pipeline, --build, --step-key, --group-key, --since,
+--until, --queue, --state
 
-With --build, --state is also applied by the server. Client-side filters and
-ordering are applied after all required cursor pages have been fetched.
+Client-side filters: --duration, and --state without --queue or --build
```
mitchbne, line 57: I think we could do state fully server side (even if we don't have queue or build?). I wonder if we could also do duration server side too?

JoeColeman95: Both true but have some quite significant drawbacks I never really explained properly tbf.

Duration can't be server side here because there's no `duration` or `startedAt` filter on the jobs, there is `createdAtFrom` and `createdAtTo`, and `finishedAt`, but that'd resolve to the nearest day rather than hourly, for example.

State is doable, but I think the better trade off here is client side as the complexity of the query would result in someone being able to get about 2 pages per 5 mins of results before they hit the default limit (graph). I'm using REST here as it's the cheaper option, which would allow users to get about 50 pages per min of results w/ the client-side filter. REST just returns a lot more per request in the case of jobs.

PR #944, opened 2026-08-12
Add SSH access for hosted jobs

jamiemonserrate:
### Description

Add `bk job ssh <job-uuid>` so users can open an interactive shell on a running macOS hosted job from their current terminal.

The command requests short-lived connection material from the Buildkite REST API and negotiates SSH in-process over an authenticated TLS WebSocket ingress to the macOS VM. It does not write the returned private key to disk or launch the system `ssh` binary.

Linux hosted jobs are intentionally out of scope. They will follow shortly

This draft depends on the macOS session contract in [buildkite/buildkite#32391](https://github.com/buildkite/buildkite/pull/32391). Until that endpoint is deployed, the command returns 404 as expected.

Context: [A-1651](https://linear.app/buildkite/issue/A-1651/support-secure-bk-job-ssh-access-for-macos-jobs)

### Changes

- Add the visible `bk job ssh <job-uuid>` command for macOS hosted jobs.
- Reuse the ingress transport and shared remote-session fields from VNC.
- Authenticate the `wss://` endpoint through its TLS certificate and hostname before exchanging SSH traffic.
- Connect with the API-provided SSH username and ephemeral private key.
- Forward stdin, stdout, stderr, terminal dimensions, and resize events.
- Restore terminal state and close the connection cleanly on interruption, including during the SSH handshake.
- Redact SSH private keys from debug HTTP response dumps.
- Cover the REST contract, real SSH over ingress, authentication, cancellation, registration, and credential redaction.

```text
Usage: bk job ssh <job-uuid>

Connect to a running hosted job over SSH.

Arguments:
  <job-uuid>    UUID of the hosted agent job
```

### Testing

- [x] Tests have run locally (`go test -p 1 ./...`)
- [x] Code is formatted (`gofmt` and `git diff --check`)
- [x] Focused race tests pass (`go test -race ./cmd/job ./pkg/cmd/factory .`)
- [x] Windows job package cross-compiles
- [x] `golangci-lint` passes
- [x] `govulncheck` passes with the pinned Go 1.26.5 toolchain
- [x] Manual hosted-job acceptance test (blocked until the `ssh-session` API endpoint is available)

### Disclosures / Credits

Implemented with Codex.

### DIFF
--- a/cmd/job/remote_session.go
+++ b/cmd/job/remote_session.go
@@ -3,7 +3,6 @@ package job
 import (
 	"context"
 	"errors"
-	"fmt"
 	"time"
 )
 
@@ -13,14 +12,14 @@ type remoteSession struct {
 	ExpiresAt   time.Time `json:"expires_at"`
 }
 
-func (s remoteSession) validate(kind string) error {
+func (s remoteSession) validate() error {
 	switch {
 	case s.Endpoint == "":
-		return fmt.Errorf("buildkite API returned %s session without an endpoint", kind)
+		return errors.New("without an endpoint")
 	case s.AccessToken == "":
-		return fmt.Errorf("buildkite API returned %s session without an access token", kind)
+		return errors.New("without an access token")
 	case s.ExpiresAt.IsZero():
-		return fmt.Errorf("buildkite API returned %s session without an expiry", kind)
+		return errors.New("without an expiry")
 	default:
 		return nil
 	}
--- a/cmd/job/remote_session.go
+++ b/cmd/job/remote_session.go
@@ -1,0 +1,36 @@
+package job
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+type remoteSession struct {
+	Endpoint    string    `json:"endpoint"`
+	AccessToken string    `json:"access_token"`
+	ExpiresAt   time.Time `json:"expires_at"`
+}
+
+func (s remoteSession) validate(kind string) error {
+	switch {
+	case s.Endpoint == "":
+		return fmt.Errorf("buildkite API returned %s session without an endpoint", kind)
+	case s.AccessToken == "":
+		return fmt.Errorf("buildkite API returned %s session without an access token", kind)
+	case s.ExpiresAt.IsZero():
+		return fmt.Errorf("buildkite API returned %s session without an expiry", kind)
+	default:
+		return nil
+	}
+}
+
+type remoteAccessToken string
+
+func (t remoteAccessToken) IssueToken(context.Context, time.Duration, bool) (string, error) {
+	if t == "" {
+		return "", errors.New("remote access token is empty")
+	}
+	return string(t), nil
+}
--- a/cmd/job/ssh.go
+++ b/cmd/job/ssh.go
@@ -50,14 +50,14 @@ func (s sshSession) validate() error {
 		return errors.New("buildkite API returned an SSH session endpoint without a hostname")
 	}
 
-	switch {
-	case s.SSH.Username == "":
+	if s.SSH.Username == "" {
 		return errors.New("buildkite API returned an SSH session without an SSH username")
-	case s.SSH.PrivateKey == "":
-		return errors.New("buildkite API returned an SSH session without an SSH private key")
-	default:
-		return nil
 	}
+	if s.SSH.PrivateKey == "" {
+		return errors.New("buildkite API returned an SSH session without an SSH private key")
+	}
+
+	return nil
 }
 
 func (c *SSHCmd) Help() string {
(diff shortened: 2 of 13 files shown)


cmd/job/ssh.go
```diff
@@ -0,0 +1,255 @@
...
+			_, _ = io.Copy(remoteStdin, streams.Stdin)
+			_ = remoteStdin.Close()
+		}()
+	}
+	if err := sshShell.Wait(); err != nil {
+		if ctx.Err() != nil {
+			return nil
+		}
+		return fmt.Errorf("wait for SSH shell: %w", err)
+	}
```
JoeColeman95, line 198: This will throw an error if you exit the session with exit/Ctrl+C which feels like an oversight? There is handling for isExpectedVSNDisconnect, but that doesn't catch this.

cmd/job/ssh.go
```diff
@@ -0,0 +1,255 @@
...
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	return c.run(ctx, sshStreams{
+		Stdin:         os.Stdin,
+		Stdout:        kongCtx.Stdout,
+		Stderr:        kongCtx.Stderr,
+		TerminalInput: os.Stdin,
+		TerminalSize:  terminalSizeFile(os.Stdin, os.Stdout),
```
JoeColeman95, line 93: On Windows `terminalSizeFile` gives us `os.Stdout` but we're gating the PTY on stdin, so an output redirect here results in an abandon

cmd/job/ssh.go
```diff
@@ -0,0 +1,255 @@
...
+			return nil
+		}
+		return fmt.Errorf("start SSH shell: %w", err)
+	}
+	if remoteStdin != nil {
+		go func() {
+			_, _ = io.Copy(remoteStdin, streams.Stdin)
+			_ = remoteStdin.Close()
+		}()
+	}
```
JoeColeman95, line 192: There's two `remoteStdin.Close()` in here which can both occur in `CloseWrite` and can result in a race condition during tear down

JoeColeman95 (dismissed):

cmd/job/ssh.go
jamiemonserrate: Thanks @JoeColeman95. This is now fixed.

cmd/job/ssh.go
jamiemonserrate: Thanks @JoeColeman95. This is now fixed.

cmd/job/ssh.go
jamiemonserrate: Thanks @JoeColeman95. This is now fixed.

cmd/job/ssh.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("buildkite API returned an SSH session with an invalid endpoint: %w", err)
+	}
+	if endpoint.Scheme != "wss" {
+		return fmt.Errorf("buildkite API returned an SSH session endpoint that must use %q, got %q", "wss", endpoint.Scheme)
+	}
+	if endpoint.Hostname() == "" {
+		return errors.New("buildkite API returned an SSH session endpoint without a hostname")
+	}
+
+	switch {
```
buildkate: Can this be if statements just to keep consistency within this validate function

jamiemonserrate: Done!

cmd/job/remote_session.go
```diff
@@ -0,0 +1,36 @@
...
+	"time"
+)
+
+type remoteSession struct {
+	Endpoint    string    `json:"endpoint"`
+	AccessToken string    `json:"access_token"`
+	ExpiresAt   time.Time `json:"expires_at"`
+}
+
+func (s remoteSession) validate(kind string) error {
```
buildkate: nit: can we just remove the argument `kind string` from this, return a plain string error, and wrap it at the call sites like 

```
  if err := s.remoteSession.validate(); err != nil {
    return fmt.Errorf("VNC session: %w", err)
}
```

Reason being, I find the current call sites read a bit odd, having to use proper grammar in the args
for example:
```
if err := s.remoteSession.validate("an SSH"); err != nil {
		return err
	}
```

cmd/job/ssh.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+func isExpectedSSHExit(ctx context.Context, err error) bool {
+	if ctx.Err() != nil {
+		return true
+	}
+
+	var exitErr *ssh.ExitError
+	if errors.As(err, &exitErr) {
```
buildkate: Am I reading this correctly and is this the desired behaviour?

If the remote command exits non-0 we won't fail the cli command?

jamiemonserrate: Damn, this was a good find! Thank you. Fixing now.

cmd/job/remote_session.go
jamiemonserrate: Fair point. Fixing.

buildkate (approved):

PR #959, opened 2026-08-24
Omit cluster secret request bodies from REST debug logs

lox:
### Description

`bk --debug` dumps REST request bodies to stderr. Cluster secret create and value-update requests contain plaintext secret material, which should never be copied into terminal or CI logs.

### Changes

- Omit request bodies entirely for cluster secret create and value-update endpoints while retaining request metadata and redacted headers.
- Preserve byte-for-byte request forwarding and continue showing bodies for unrelated endpoints.
- Recognize malformed secret paths, including extra slashes and URL delimiters, so invalid IDs cannot expose a body before the API rejects the request.
- Leave response debug dumps unchanged because the API does not return secret values.

### Testing

### Disclosures / Credits

Amp implemented and tested this change under my direction.

### DIFF
--- a/pkg/cmd/factory/factory.go
+++ b/pkg/cmd/factory/factory.go
@@ -159,14 +159,13 @@ func redactRequestDump(req *http.Request, dump string) string {
 
 func isClusterSecretValueRequest(req *http.Request) bool {
 	path := strings.TrimSuffix(req.URL.Path, "/")
-	parts := strings.Split(strings.Trim(path, "/"), "/")
-	if req.Method == http.MethodPost && len(parts) >= 3 {
-		return parts[len(parts)-3] == "clusters" && parts[len(parts)-1] == "secrets"
+	if !strings.Contains(path, "/clusters/") {
+		return false
 	}
-	if req.Method == http.MethodPut && len(parts) >= 5 {
-		return parts[len(parts)-5] == "clusters" && parts[len(parts)-3] == "secrets" && parts[len(parts)-1] == "value"
+	if req.Method == http.MethodPost {
+		return strings.HasSuffix(path, "/secrets")
 	}
-	return false
+	return req.Method == http.MethodPut && strings.Contains(path, "/secrets/") && strings.HasSuffix(path, "/value")
 }
 
 // sensitiveBodyPatterns matches token values in form-encoded request bodies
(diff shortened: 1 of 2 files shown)


pkg/cmd/factory/factory.go
```diff
@@ -124,6 +125,50 @@ func (d *debugTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	if err != nil {
+		return redactBody(dump[:bodyStart] + "[REDACTED]")
+	}
+
+	return redactBody(dump[:bodyStart] + string(redacted))
+}
+
+func isClusterSecretValueRequest(req *http.Request) bool {
+	path := strings.TrimSuffix(req.URL.Path, "/")
+	parts := strings.Split(strings.Trim(path, "/"), "/")
```
SorchaAbel: @lox looks like we don’t validate the cluster UUID, passing something like 
--cluster-uuid 'cluster-1/' 
creates 
/clusters/cluster-1//secrets. 
The extra slash means this isn’t recognised as a secret request, so the value isn’t redacted.

lox: Fixed in 02a1993: secret request detection now fails closed for malformed cluster and secret IDs that introduce extra path segments, with regression coverage for the reported double-slash path.

lox: Fixed in 37149dd: detection now reconstructs the parsed URL target from path, query, and fragment before matching secret endpoints. Regression tests cover `?` in a cluster ID and `#` in a secret ID, alongside the existing extra-slash case.

pkg/cmd/factory/factory.go
```diff
@@ -124,6 +131,24 @@ func (d *debugTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return resp, nil
 }
 
+func isClusterSecretValueRequest(req *http.Request) bool {
+	target := req.URL.Path
+	if req.URL.RawQuery != "" {
```
SorchaAbel, line 141: so I _could_ be wrong but if the request URL is 
`/secrets?somequerystring=something`, 
appending `RawQuery` makes `HasSuffix(target, "/secrets")` return false, so the request body is logged without redacting the secret value 😬 .

lox: Fixed in d57c504: endpoint detection now matches `URL.Path` before considering query or fragment data, so a normal query string cannot disable body omission. Reconstructing the full parsed target remains only the fallback for malformed IDs that move the endpoint suffix across a URL delimiter. Added regression coverage for `/secrets?somequerystring=something`.

mcncl (approved):

PR #975, opened 2026-09-02
Allow device login to preselect an organization

sj26:
### Description

`bk auth login --device` currently rejects `--org`, so multi-organization users must select the intended organization manually in the browser.

Allow device login to send the same organization slug or UUID hint that browser-based login already supports:

```bash
bk auth login --device --org my-org
```

Requires the additive device authorization endpoint support in https://github.com/buildkite/buildkite/pull/33566. Older endpoints safely ignore the hint. User documentation is in https://github.com/buildkite/docs/pull/3159.

### Changes

- Allow `--device` and `--org` to be used together.
- Send `organization` or `organization_uuid` with the device authorization request.
- Prefer UUID when both values are present and update the device-login help example.

### Testing

- [x] Affected OAuth and auth command tests pass locally.
- [x] Code is formatted (with `mise run format`).
- [x] `mise run lint` passes.
- [ ] Full `go test ./...` is blocked locally by the existing `TestCmdSwitch` failures, which reproduce when that untouched test runs alone.

### Disclosures / Credits

Amp investigated, implemented, reviewed, and tested this change under my direction.

### DIFF
--- a/cmd/auth/login.go
+++ b/cmd/auth/login.go
@@ -59,8 +59,8 @@ Examples:
   # Login non-interactively with an API token
   $ bk auth login --org my-org --token my-token
 
-  # Login on a headless machine or remote shell
-  $ bk auth login --device
+  # Login to a specific organization on a headless machine or remote shell
+  $ bk auth login --device --org my-org
 
   # Login on a headless Linux host using an in-memory /dev/shm credential store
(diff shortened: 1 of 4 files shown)


sj26:
@buildsworth-bk review, approve L2

PriyaSudip (approved):

PR #982, opened 2026-09-07
Prevent OAuth refresh when credentials cannot be persisted

### LINKED ISSUE #981: OAuth refresh loses auth when macOS Keychain is readable but not writable
### Description

I hit this with `bk` 3.56.0 on macOS 26.5.1 while running it through the Codex agent harness. The harness process can read the login Keychain but cannot write to it.

`bk` works until the one-hour OAuth access token expires. Refresh succeeds server-side, but persisting the rotated refresh token fails:

```text
credential persistence failed: failed to store rotated refresh token: exit status 161
```

Subsequent commands then fail with:

```text
token refresh error: invalid_grant - Invalid refresh token
401 Authentication required
```

This also reproduces with a single sequential `bk` process, so concurrent watchers are not required.

### Diagnosis

The underlying `/usr/bin/security -i` write reports:

```text
SecKeychainItemCreateFromContent: UNIX[Operation not permitted]
add-generic-password: returned 100001
```

`100001` becomes process exit status `161`. The same Keychain write and automatic `bk` refresh succeed outside the sandbox.

The refresh exchange happens before the replacement refresh token is persisted. Once the write fails, the stored refresh token has already been rotated and is unusable.

### Expected behavior

Before consuming a rotating refresh token, `bk` should verify that the selected `auto` credential store is writable, provide a writable fallback, or return an actionable Keychain permission error without invalidating the stored session.

### Relevant PRs

#841
#852

Spikhalskiy:
Fixes #981.

OAuth refresh tokens rotate when exchanged. The CLI currently exchanges the token before discovering that its credential store cannot persist the replacement, which leaves the stored refresh token invalid for subsequent commands.

This change performs a write/read/delete probe using a unique, non-secret credential immediately before the exchange. A write failure aborts before the server can rotate the real token. In `auto` mode, the same probe also selects the SHM fallback when it is available; the existing post-exchange persistence checks remain in place for failures that occur after the probe.

Tests cover an unwritable store making zero OAuth requests, direct keyring write failures, SHM fallback selection, and the existing post-exchange failure path.


mcncl:
@Spikhalskiy thanks for opening this, I'll take a look. A workaround/fix to unblock you here might be;

```
security unlock-keychain ~/Library/Keychains/login.keychain-db
mise reshim
bk auth login --org ORG-SLUG
bk auth status
```

This assumes you're using `mise` for binary management, so will need adjustment for your own setup.

mcncl (approved):

Spikhalskiy:
@mcncl thank you for looking into this! Yeah, even simpler workaround would be just to generate a token with a long expiry and wire it as `BUILDKITE_API_TOKEN` env var.

PR #997, opened 2026-09-21
Preserve broken job reasons in build view JSON

bbrown-oai:
## Motivation

`bk build view --json` drops each job's `broken_reason` because the SDK's `Job` type does not decode it. Scripts cannot distinguish intentionally omitted jobs from other jobs that never started.

## Summary

Decode build-view jobs through a typed response that includes `broken_reason`. Keep the SDK client for authentication and requests, and preserve job-state filtering, summary requests, text output, artifacts, and annotations.

Omit structured build, pipeline, and step environments, agent access tokens, and secret pipeline webhook URLs from build-view output. YAML retains its `build.jobs` structure; the additional reason field is in JSON.

The SDK fix in buildkite/go-buildkite#381 is still open. This change works with the CLI's current released SDK dependency.

## Validation

- Built with the repository-pinned Go 1.27.1; repository format and lint checks passed.
- Direct serialization tests cover reason codes, free-text reasons, null/absent reasons, retained build data, environment/credential omission, and YAML job placement. Existing non-mock summary and option tests passed.
- Authenticated read-only comparisons against two real builds preserved all 11 non-empty API reasons. The baseline omitted those reasons; the patched output had no structured environment or agent-token fields.
- Full text and summary JSON matched the baseline byte for byte. A live `--job-states broken` request returned the expected jobs and reasons.

Local validation used real API reads and tests without HTTP mocks. The full upstream suite remains for CI.


bbrown-oai:
@codex review

bbrown-oai:
@codex security review

